PR #8, opened 2022-09-19
Op datastore abi 1

sydhds:


assembly/std/index.ts
```diff
@@ -102,6 +102,94 @@ export function balanceOf(address: string): u64 {
...
+ * @return {Array<StaticArray<u8>} - a list of key (e.g. a list of bytearray)
+ */
+export function getOpKeys(): Array<StaticArray<u8>> {
+  let keys_ser = env.getOpKeys();
+  return derOpKeys(keys_ser);
+}
+
+export function derOpKeys(keys_ser: StaticArray<u8>): Array<StaticArray<u8>> {
+
+  let MAX_DATASTORE_ENTRY_COUNT = u16.MAX_VALUE;
```
damip, line 143: maybe it's worth defining a constants.ts for this (and error codes in the future)

assembly/std/index.ts
```diff
@@ -102,6 +102,94 @@ export function balanceOf(address: string): u64 {
...
+  let default_res = new Array<StaticArray<u8>>();
+  let keys_der = new Array<StaticArray<u8>>();
+  // Datastore deserialization
+  // Format is: L (u32); V1_L (u8); V1 data (u8*V1_L); ...
+  // TODO: test keys_ser initial len
+  // u8 * 4 (LE) => u32
+  let entry_count: u32 = keys_ser[0] << 24 + keys_ser[1] << 16 + keys_ser[2] << 8 + keys_ser[3];
```
damip, line 150: Would https://www.assemblyscript.org/stdlib/dataview.html be cleaner ?

assembly/std/index.ts
```diff
@@ -102,6 +102,94 @@ export function balanceOf(address: string): u64 {
...
+
+  if (entry_count == 0 || entry_count > MAX_DATASTORE_ENTRY_COUNT) {
```
damip, line 152: in the case `entry_count > MAX_DATASTORE_ENTRY_COUNT` for now we should `throw` (to avoid silent unexpected states). In the future it could be an error return

damip (commented):
Good work !
Just a few points to address

assembly/std/index.ts
sydhds: Done

assembly/std/index.ts
sydhds: Done

assembly/std/index.ts
```diff
@@ -102,6 +102,94 @@ export function balanceOf(address: string): u64 {
...
+    return keys_der;
+  }
+
+  let entry_pushed: u32 = 0;
+  let i = 4;
+  while(i < keys_ser.length) {
```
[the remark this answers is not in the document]
sydhds: Done for the prealloc

assembly/std/index.ts
sydhds: Problem is that as-pect fails to catch the exception for weird reason... Should I already implement the error return or a Result like return ?

damip: no, then let's keep the error slient.
Note that this check shouldn't be useful in the first place: we are supposed to trust what the ABI returns. If it sends something invalid, it's normal to panic

assembly/std/index.ts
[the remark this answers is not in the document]
damip: Internally an array is a pair (length_of_the_array, offset_of_the_first_element_of_the_array_in_linear_memory).

For example every TypedArray has the following method:
```
function subarray(start?: i32, end?: i32): TypedArray

Returns a new view on the array's backing buffer from begin inclusive to end exclusive relative to this array. If omitted, end defaults to the end of the array. Does not copy.
```
Special emphasis on `Does not copy.`.

This means that we can just deserialize the lengths and return an array of arrays representing the list of keys, without ever copying a single slice representing a key.

assembly/std/index.ts
```diff
@@ -102,6 +103,103 @@ export function balanceOf(address: string): u64 {
...
+  // we can safely cast from u8 to bool
+  return bool(result[0]);
+}
+
+/*
+ * Get data associated with the given key from datastore
+ *
+ * @param {StaticArray<u8>} key
+ *
+ * @return {StaticArray<u8>} - data as a byte array
```
damip, line 126: the documentation (teyturn type) of this function (and others) needs to be updated

assembly/std/index.ts
```diff
@@ -102,6 +103,103 @@ export function balanceOf(address: string): u64 {
...
+export function getOpKeys(): Array<Uint8Array> {
+  let buf = env.getOpKeys();
+  let keys_ser: Uint8Array = Uint8Array.wrap(buf);
+  return derOpKeys(keys_ser);
+}
+
+/*
+ * Internal function - used by getOpKeys
+ */
+export function derOpKeys(keys_ser: Uint8Array): Array<Uint8Array> {
```
damip, line 147: I still believe the implem of derOpKeys is a bit too complicated. Input from the ABI towards the smart contract can be trusted to minimize gas usage and code complexity. The other way around is not true however: data sent from smart contracts to the rust abi should be checked because it is non-validated user input.


Here is a recommendation (also avoids creating default_res when it is not needed):

```
export function derOpKeys(keys_ser: Uint8Array): Array<Uint8Array> {
  if (keys_ser.length == 0) {
    return new Array<Uint8Array>();
  }

  // Datastore key list deserialization
  // Format is: L (u32); V1_L (u8); V1 data (u8*V1_L); ...
  // u8 * 4 (LE) => u32
  let dv = new DataView(keys_ser.buffer, 0, 4);
  let entry_count = dv.getUint32(0, true /* littleEndian */);
  
  let cursor = 4;
  let keys_der = new Array<Uint8Array>(entry_count);
  for (let i = 0; i < entry_count; i++) {
    let end = cursor + keys_ser[cursor] + 1;
    keys_der[entry_pushed] = keys_ser.subarray(cursor + 1, end); // no copy here
    cursor = end;
  }
    
  return keys_der;
}
```

assembly/std/index.ts
sydhds: done

assembly/std/index.ts
sydhds: Ok for trusting input from the ABI but this is not bullet proof. Any unchecked (or without enough review) modification (e.g from a newcomer) can cause an issue in deserialization. Moreover because we remove the checks, reporting the proper problem will not be possible.

Anyway it's done and unit tests about these edge cases & invalid format have been removed.

damip (approved):
Very good work !

PR #15, opened 2022-10-14
add astester.imports.js

SebastienValla:
add a envy.imports.js to mock VM and make unit tests with for the moment : 
generateEvent
setOf
getOf


astester.imports.js
```diff
@@ -0,0 +1,39 @@
...
+  return {
+    env: {
+      memory,
+      abort(msg, file, line, col) {
+      getString(msg)
+
+       console.log("error", text)
+      },
+      log(ptr) {
+       console.log("log", text)
```
gregLibert, line 12: How can that work ? What is this magic ?

astester.imports.js
```diff
@@ -0,0 +1,39 @@
...
+      },
+  },
+  massa: {
+      memory,
```
gregLibert, line 16: not sure you need that.

astester.imports.js
```diff
@@ -0,0 +1,39 @@
...
+      assembly_script_generate_event(string) {
+       console.log("event", string)
```
gregLibert, line 18: not sure you can do that. I think you will have a pointer. You should have a look at what Max did for the playground.

astester.imports.js
```diff
@@ -0,0 +1,39 @@
...
+      },
+      assembly_script_set_data_for(address, key, value) {
```
gregLibert, line 20: Same, if it's string, you will have pointer

astester.imports.js
```diff
@@ -0,0 +1,39 @@
...
+      },
+      assembly_script_get_data_for(address, key) {
+          let value = "";
+          if (Storage.has(address)) {
+              const addressStorage = Storage.get(address);
+              if (addressStorage.has(key)) {
+                  value = addressStorage.get(key);
+              }
+          }
+          return value;
```
gregLibert, line 35: Same, not sure it will work. You should return a pointer from distant memory after setting the string there.

gregLibert (changes requested):

astester.imports.js
SebastienValla: yes indeed it was ! issue solve with getString

astester.imports.js
SebastienValla: needed to include it in astester/cli

astester.imports.js
SebastienValla: indeed, problem solved

gregLibert (approved):

peterjah (approved):

PR #23, opened 2022-10-26
Add Args class

Thykof:
This branch starts on the the `fmt-lint-hook-ci` branch

### DIFF
--- a/assembly/std/__tests__/arguments.spec.ts
+++ b/assembly/std/__tests__/arguments.spec.ts
@@ -9,7 +9,10 @@ describe('Args tests', () => {
     // Create an argument class instance
     const args1 = new Args();
     // add some arguments
-    args1.add(new Address(ADDR0)).add(new Address(ADDR1)).add(97);
+    args1
+      .add(new Address(ADDR0))
+      .add(new Address(ADDR1))
+      .add(97 as u64);
 
     // use serialize to get the byte string
     const byteString = args1.serialize();
@@ -28,13 +31,13 @@ describe('Args tests', () => {
 
   it('With a number and an address', () => {
     const args1 = new Args();
-    args1.add(97).add(new Address(ADDR0));
+    args1.add(97 as u32).add(new Address(ADDR0));
 
-    expect(args1.nextU64()).toBe(97);
+    expect(args1.nextU32()).toBe(97 as u32);
     expect(args1.nextAddress().toByteString()).toBe(ADDR0);
 
     const args2 = new Args(args1.serialize());
-    expect(args2.nextU64()).toBe(97);
+    expect(args2.nextU32()).toBe(97 as u32);
     expect(args2.nextAddress().toByteString()).toBe(ADDR0);
   });
 
--- a/assembly/std/__tests__/arguments.spec.ts
+++ b/assembly/std/__tests__/arguments.spec.ts
@@ -92,6 +92,16 @@ describe('Args tests', () => {
     expect(args2.nextI64()).toBe(113);
   });
 
+  it('With a big string', () => {
+    const args1 = new Args();
+    args1.add('a'.repeat(257));
+
+    expect(args1.nextString()).toBe('a'.repeat(257));
+
+    const args2 = new Args(args1.serialize());
+    expect(args2.nextString()).toBe('a'.repeat(257));
+  });
+
   it('With string and u64', () => {
     const args1 = new Args();
     args1
@@ -114,15 +124,15 @@ describe('Args tests', () => {
     args1
       .add(0 as u64)
       .add('my string')
-      .add(0xffffffffffffffff as u64);
+      .add(u64.MAX_VALUE as u64);
 
     expect(args1.nextU64()).toBe(0);
     expect(args1.nextString()).toBe('my string');
-    expect(args1.nextU64()).toBe(0xffffffffffffffff);
+    expect(args1.nextU64()).toBe(u64.MAX_VALUE);
 
     const args2 = new Args(args1.serialize());
     expect(args2.nextU64()).toBe(0);
     expect(args2.nextString()).toBe('my string');
-    expect(args2.nextU64()).toBe(0xffffffffffffffff);
+    expect(args2.nextU64()).toBe(u64.MAX_VALUE);
   });
 });
--- a/assembly/std/__tests__/arguments.spec.ts
+++ b/assembly/std/__tests__/arguments.spec.ts
@@ -6,12 +6,23 @@ const ADDR1 = 'A1Czd9sRp3mt2KU9QBEEZPsYxRq9TisMs1KnV4JYCe7Z4AAVinq';
 
 describe('Args tests', () => {
   it('With 2 addresses and a number', () => {
+    // Create an argument class instance
     const args1 = new Args();
+    // add some arguments
     args1.add(new Address(ADDR0)).add(new Address(ADDR1)).add(97);
 
-    const args2 = new Args(args1.serialize());
-    expect(args2.nextAddress()._value).toBe(ADDR0);
-    expect(args2.nextAddress()._value).toBe(ADDR1);
+    // use serialize to get the byte string
+    const byteString = args1.serialize();
+    // you can then use it in the call function:
+    // env.call(at.toByteString(), functionName, byteString, coins);
+
+    // create an argument class with the byte string
+    const args2 = new Args(byteString);
+    // assert that the first address is same we provide
+    // in the first call to add function
+    expect(args2.nextAddress().toByteString()).toBe(ADDR0);
+    // and so on with the 2 following arguments
+    expect(args2.nextAddress().toByteString()).toBe(ADDR1);
     expect(args2.nextU64()).toBe(97);
   });
 
@@ -20,11 +31,11 @@ describe('Args tests', () => {
     args1.add(97).add(new Address(ADDR0));
 
     expect(args1.nextU64()).toBe(97);
-    expect(args1.nextAddress()._value).toBe(ADDR0);
+    expect(args1.nextAddress().toByteString()).toBe(ADDR0);
 
     const args2 = new Args(args1.serialize());
     expect(args2.nextU64()).toBe(97);
-    expect(args2.nextAddress()._value).toBe(ADDR0);
+    expect(args2.nextAddress().toByteString()).toBe(ADDR0);
   });
 
   it('With Address and i64', () => {
@@ -35,12 +46,12 @@ describe('Args tests', () => {
       .add(113 as i64);
 
     expect(args1.nextI64()).toBe(97);
-    expect(args1.nextAddress()._value).toBe(ADDR0);
+    expect(args1.nextAddress().toByteString()).toBe(ADDR0);
     expect(args1.nextI64()).toBe(113);
 
     const args2 = new Args(args1.serialize());
     expect(args2.nextI64()).toBe(97);
-    expect(args2.nextAddress()._value).toBe(ADDR0);
+    expect(args2.nextAddress().toByteString()).toBe(ADDR0);
     expect(args2.nextI64()).toBe(113);
   });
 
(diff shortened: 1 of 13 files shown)


SebastienValla:
still in draft ?

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+   * @param {u8} offset
+   * @return {i64}
+   */
+  private toI64(byteArray: Uint8Array, offset: u8 = 0): i64 {
+    if (byteArray.length - offset < 8) {
+      return <i64>NaN;
+    }
+
+    let x: i64 = 0;
+    x = (x | byteArray[offset + 7]) << 8;
```
peterjah, line 155: So the toU64() does the same thing as toI64() ?
I think you should add Unit tests to test these functions.
and if its really the same processing inside, factorize the code.

Personnally i would to it like this (not tested)
x= x | byteArray[offset] as u64
x= x | ( byteArray[offset + 1] as u64 >> 1)
x= x | ( byteArray[offset + 2) as u64 >> 2)

peterjah (changes requested):

assembly/std/arguments.ts
peterjah: I understood the logic 🙌
and it seems this code is already tested in some way
Code factorization will be nice but not mandatory.
So i will cancel my change request !

peterjah (approved):

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+   * @param {u8} offset
+   * @return {u64}
+   */
+  private toU64(byteArray: Uint8Array, offset: u8 = 0): u64 {
+    if (byteArray.length - offset < 8) {
+      return <u64>NaN;
+    }
+
+    let x: u64 = 0;
+    x = (x | byteArray[offset + 7]) << 8;
```
peterjah, line 133: for (let i = 7; i >= 0; --i) {
  x = (x | byteArray[offset + i]) << 8;
}
:smiley:

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+export class Args {
+  private offset: i64 = 0;
+  private argsString: string = '';
+
+  /**
+   *
+   * @param {string} argsString
+   * @param {Array<string>} types
+   */
+  constructor(argsString: string = '') {
```
gregLibert, line 16: This argument is misleading : you reuse the name of the package and the name of the type.

Using something like `serialized` will make the constructor clearer:

```typescript
var args = new Args(serialized);
```

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
+import {Address} from './address';
+import {ByteArray} from '@massalabs/as/assembly/byteArray';
+
+/**
+ * Args for remote function call
```
gregLibert, line 5: You should explain in few words what this class is doing.

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+   *
+   * @param {string} argsString
+   * @param {Array<string>} types
+   */
+  constructor(argsString: string = '') {
+    this.argsString = argsString;
+  }
+
+  /**
+   *
```
gregLibert, line 21: You should add documentation to all function explaining what it does.

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  // getters
+
+  /**
+   *
+   * @return {Address} the address
+   */
+  nextAddress(): Address {
+    const address = new Address();
+    this.offset = address.fromStringSegment(
```
gregLibert, line 36: You are enforcing a bad practice: serialization/deserialization from/to string. The good practice is to use byte arrays to do so.

It will not be possible to add all objects here, so you should consider this function as a documented template on how to deserialize any object from Args. And therefore, document this function as such.

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+    return address;
+  }
+
+  /**
+   *
+   * @return {string} the string
+   */
+  nextString(): string {
+    let offset: i32 = this.offset as i32;
+    const length = u8(this.argsString.codePointAt(offset));
```
gregLibert, line 49: You should add a warning in the documentation of this function that the max size of a string is 255 characters.

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+    const result = this.argsString.slice(offset + 1, offset + length + 1);
```
gregLibert, line 50: Why the last +1 ?
Using a variable such as `end` will make it clearer that the second index is not the last element, but the first after (hence the classical end denomination) and it will avoid recopying 'offset + length + 1' twice :-)

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+    const amount = this.toU64(byteArray, this.offset as u8);
+    this.offset += 8;
+    return amount;
+  }
+
+  /**
+   * @return {i64}
+   */
+  nextI64(): i64 {
+    const byteArray = this.fromByteString(this.argsString);
```
gregLibert, line 69: duplicate of nextU64. Can you simplify that using typechange<T> ?

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   *
+   * @param {T} arg the argument to add
+   *
+   * @return {Args} the modified Arg instance
+   */
+  add<T>(arg: T): Args {
+    if (arg instanceof Address) {
+      this.argsString = this.argsString.concat(arg.toStringSegment());
```
gregLibert, line 85: Same as above, you should use Byte array exported function of external object to follow good practice.

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+    } else if (arg instanceof String || typeof arg == 'string') {
```
gregLibert, line 86: String hasn't a typeof 'string' ?

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+    } else if (arg instanceof u64) {
+      this.argsString = this.argsString.concat(
+        ByteArray.fromU64(arg as u64).toByteString(),
+      );
+    } else if (arg instanceof i64 || typeof arg == 'number') {
+      this.argsString = this.argsString.concat(
+        ByteArray.fromI64(arg as i64).toByteString(),
+      );
+    } else {
+      throw new Error('unsupported type: ' + typeof arg);
```
gregLibert, line 100: If you throw, you stop the execution of the SC. There is no catch possibility right now in AS.

gregLibert (changes requested):

assembly/std/arguments.ts
Thykof: right

assembly/std/__tests__/arguments.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+const ADDR0 = 'A12cap3Gd1bDwVaY7LkPAj7GAayKueNq6ebaPELBfBhQLqR3R7rg';
+const ADDR1 = 'A1Czd9sRp3mt2KU9QBEEZPsYxRq9TisMs1KnV4JYCe7Z4AAVinq';
+
+describe('Args tests', () => {
+  it('With 2 addresses and a number', () => {
+    const args1 = new Args();
+    args1.add(new Address(ADDR0)).add(new Address(ADDR1)).add(97);
+
+    const args2 = new Args(args1.serialize());
+    expect(args2.nextAddress()._value).toBe(ADDR0);
```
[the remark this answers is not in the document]
Thykof: you're right, I will use toByteString

assembly/std/__tests__/arguments.spec.ts
```diff
@@ -0,0 +1,114 @@
+import {Address} from '../address';
+import {Args} from '../arguments';
+
+const ADDR0 = 'A12cap3Gd1bDwVaY7LkPAj7GAayKueNq6ebaPELBfBhQLqR3R7rg';
+const ADDR1 = 'A1Czd9sRp3mt2KU9QBEEZPsYxRq9TisMs1KnV4JYCe7Z4AAVinq';
+
+describe('Args tests', () => {
+  it('With 2 addresses and a number', () => {
```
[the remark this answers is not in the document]
Thykof: new issue https://github.com/massalabs/massa-as-sdk/issues/34

assembly/std/arguments.ts
Thykof: okay I'll try something

assembly/std/arguments.ts
Thykof: good catch

assembly/std/arguments.ts
Thykof: okay so I've removed the throw, so no warnings, just do nothing in that case

assembly/std/__tests__/arguments.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    expect(args2.nextString()).toBe('my string');
+    expect(args2.nextU64()).toBe(11356323656733);
+  });
+
+  it('With string and 0 and max number', () => {
+    const args1 = new Args();
+    args1
+      .add(0 as u64)
+      .add('my string')
+      .add(0xffffffffffffffff as u64);
```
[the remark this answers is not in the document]
Thykof: good catch

assembly/std/arguments.ts
Thykof: done

assembly/std/arguments.ts
Thykof: done

assembly/std/arguments.ts
Thykof: done

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  // Utils
+
+  /**
+   *
+   * @param {string} byteString
+   * @return {Uint8Array}
+   */
+  private fromByteString(byteString: string): Uint8Array {
+    const byteArray = new Uint8Array(byteString.length);
```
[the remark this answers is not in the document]
Thykof: done

assembly/std/arguments.ts
```diff
@@ -0,0 +1,165 @@
...
+   * @param {u8} offset
+   * @return {i64}
+   */
+  private toI64(byteArray: Uint8Array, offset: u8 = 0): i64 {
+    if (byteArray.length - offset < 8) {
+      return <i64>NaN;
+    }
+
+    let x: i64 = 0;
+    x = (x | byteArray[offset + 7]) << 8;
```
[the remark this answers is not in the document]
Thykof: done

assembly/std/arguments.ts
Thykof: I've created an issue #35

assembly/std/arguments.ts
Thykof: I've created an issue #35

assembly/std/arguments.ts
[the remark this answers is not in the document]
gregLibert: LGTM, you should create an issue for native missing type.

assembly/std/arguments.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  /**
+   * Returns the deserialized number as u64.
+   *
+   * @return {u64}
+   */
+  nextU64(): u64 {
+    const byteArray = this.fromByteString(this.serialized);
+    const amount = this.toU64(byteArray, this.offset as u8);
```
gregLibert, line 76: amout is a bad copy/paste I think.
```suggestion
    const value = this.toU64(byteArray, this.offset as u8);
```

assembly/token/impl.ts
```diff
@@ -84,12 +84,13 @@ export function decimals(_: string): string {
 /**
  * Returns the balance of an account.
  *
- * @param {string} args - byte string containing an owner's account (Address).
+ * @param {string} argsString - byte string containing an owner's account (Address).
```
gregLibert, line 87:
```suggestion
 * @param {string} stringifyArgs - Args object serialized as a string containing an owner's account (Address).
```

gregLibert (changes requested):

assembly/token/impl.ts
Thykof: nice, I have rename the variables and rewrote the comments

assembly/std/arguments.ts
Thykof: indeed, I have renamed all `amount` into `value`

gregLibert (approved):

assembly/std/arguments.ts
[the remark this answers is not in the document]
Thykof: https://github.com/massalabs/massa-as-sdk/issues/63

PR #26, opened 2022-10-27
added some hashmap extra functonalities

BatiGencho:


gregLibert (approved):
🤷 LGTM

assembly/collections/__tests__/persistentMap.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { PersistentMap } from "../persistentMap";
+
+describe("Persistent Map tests", () => {
+  it("empty map", () => {
+    const map = new PersistentMap<string, string>("my_map");
+    assert<bool>(map.size() === 0, "empty map must have 0 size");
```
Thykof, line 6: we are now using as-pect test framework, it is more convenient to right test, see the doc here: https://as-pect.gitbook.io/as-pect/

it's ok for this PR if we stay like this with asserts

assembly/collections/persistentMap.ts
```diff
@@ -170,19 +206,26 @@ export class PersistentMap<K, V> {
    * @param {V} value The new value of the element.
    */
   set(key: K, value: V): void {
+    // assert map size wont overflow
+    assert(this._size < Usize.MAX_VALUE);
```
Thykof, line 210: this library should not panic, because there is not try/catch in assemblyscript. So rather return an "error message", we chose to implement the `Result` class in `as-types` package, you can have a look here https://github.com/massalabs/as/blob/main/packages/as-types/assembly/result.ts

it's ok for this PR to leave it like this, but we should at least add a todo comment for using `Result` and/or specify in the jsdoc that this function may panic.

Thykof (approved):

PR #27, opened 2022-11-02
Feature/token mint

BatiGencho:


assembly/token/impl.ts
```diff
@@ -8,6 +9,38 @@ import {ByteArray} from '@massalabs/as/assembly';
...
+  return false;
+}
+
+/**
+ * Sets the token owner
+ * @param {string} _args - ?
+ */
+export function setTokenOwner(_args: string): void {
+  // Check that the caller is the token owner.
+  // For initial calls, there is no owner and the check will pass.
```
peterjah, line 39: This is dangerous. your first call to setTokenOwner could be frontrunned !
i think token owner should be initialized with the token creation caller address

Thykof:
this will be done in https://github.com/massalabs/massa-sc-contracts/pull/18

PR #28, opened 2022-11-07
18 add a basic non fungible implementation

SebastienValla:
Create a first version of non fungible token implementation with:

find which implementation we should get inspired from : erc721 or 1155
create both implementation and wrapper side

add the most important unit test

Before writing code, result from the research step should be shared through comments to include most of us in the design step.

ADR for the Standard inspiration :
https://www.notion.so/massa-innoteam/NFT-Standard-choice-659d46d5a11c4c9abf0c5c4f912689e6


Thykof:
Tests run only for old as-pect test suite because new test written for envy (test for NFT) can't be executed for now in this repo.

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Return the token URI (external link written in NFT where pictures or others a stored)
+ * @param {string} tokenId
+ * @return {string}
+ */
+export function tokenURI(tokenId: string): string {
+  if (Storage.has(baseURIKey)) {
+    return Storage.get(baseURIKey) + tokenId;
+  } else {
+    generateEvent(`NFT not setted yet`);
```
Trottoir, line 93: Do we want to emit an event on a `readOnly` function ?

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Return base URI
+ * @param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
+ * @return {string}
+ */
+export function baseURI(_: string): string {
+  if (Storage.has(baseURIKey)) {
+    return Storage.get(baseURIKey);
+  } else {
+    generateEvent(`NFT not setted yet`);
```
Trottoir, line 107: Do we want to emit an event on a `readOnly` function ?

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Return the current counter, if 10 NFT minted, returns '10'.
+ *  @param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
+ *  @return {string}
+ */
+export function currentSupply(_: string): string {
+  if (Storage.has(counterKey)) {
+    return Storage.get(counterKey);
+  } else {
+    generateEvent(`NFT not setted yet`);
```
Trottoir, line 130: Do we want to emit an event on a `readOnly` function ?

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * Return the tokenId's owner
+ * @param {string} tokenId
+ * @return {string}
+ */
+export function ownerOf(tokenId: string): string {
+  if (Storage.has(ownerTokenKey + tokenId.toString())) {
+    return Storage.get(ownerTokenKey + tokenId.toString());
+  } else {
+    generateEvent(`NFT not minted yet`);
```
Trottoir, line 145: Do we want to emit an event on a `readOnly` function ?

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/**
+ * Increment the NFT counter
+ * @param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
+ * @return {string}
+ */
+function _increment(_: string): string {
+  const add1: u32 = 1;
+  const incr = u32(parseInt(Storage.get(counterKey))) + add1;
```
Trottoir, line 181:
```suggestion
  const incr = u32(parseInt(Storage.get(counterKey))) ++;
```
This'd be clearer I think

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/**
+ *
+ * Return true if the caller is the creator of the SC
+ *   @param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
+ *   @return {bool}
+ */
+function _onlyOwner(_: string): bool {
+  return Context.caller().toByteString() == Storage.get(ownerKey);
```
Trottoir, line 193: I like this <3

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,226 @@
...
+export function transfer(args: string): string {
+  const toAddress = new Address();
+  const offset = toAddress.fromStringSegment(args);
+
+  const tokenId: u64 = ByteArray.fromByteString(args.substr(offset, 8)).toU64();
+  if (!Storage.has(ownerTokenKey + tokenId.toString())) {
+    generateEvent(`token ${tokenId.toString()} not yet minted`);
+    return '';
+  }
+  if (ownerOf(tokenId.toString()) == Context.caller().toByteString()) {
```
Trottoir, line 217: We can maybe create a `modifier` ( I use the Solidity term ) just as you made with the onlyOwner
But this time smth like `onlyTokenOwner` ? Will be smth that we'll use often on that contract with the allowance for instance

assembly/NFT/NFTWrapper.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+    *Transfer a choosen token from the caller to the to Address
+    check first the caller owns the token 
+    * @param {Address} to
+    * @param {u64} tokenId
+    * @return {void}
+    */
+  tranfer(to: Address, tokenId: u64): string {
```
Trottoir, line 115:
```suggestion
  transfer(to: Address, tokenId: u64): string {
```

Trottoir (changes requested):

assembly/NFT/NFT.ts
SebastienValla: done !

assembly/NFT/NFT.ts
Thykof: I agree

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+  return ownerOf(tokenId.toString()) == Context.caller().toByteString();
+}
+
+// ==================================================== //
+// ====                 TRANSFER                   ==== //
+// ==================================================== //
+
+/**
+ *Transfer a choosen token from the caller to the to Address
+ check first the caller owns the token and if token minted
```
Thykof:
```suggestion
 * Transfer a chosen token from the caller to the to Address.
 * Check first the caller owns the token and if token minted.
 *
```

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+ * @param {string} args - byte string with the following format:
+ * - the recipient's account (address)
+ * - the tokenID (u64).
+ * @return {string}
+ */
+export function transfer(args: string): string {
+  const toAddress = new Address();
+  const offset = toAddress.fromStringSegment(args);
+
+  const tokenId: u64 = ByteArray.fromByteString(args.substr(offset, 8)).toU64();
```
Thykof:
```suggestion
  const tokenId: u64 = ByteArray.fromByteString(args.substr(offset, 8)).toU64();

```

assembly/NFT/NFTWrapper.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * This class can be used to wrap a smart contract implementing
+ * Massa standard NFT.
+ * All the serialization/deserialization will handled here.
+ *
+ * ```assemblyscript
+ *  ...
+ *  const coin = new NFTWrapper(sc_address);
+ *  const coinName = coin.Name();
+ * ...
```
Thykof:
```suggestion
 *  ...
```

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+    Storage.set(baseURIKey, newBaseURI);
+    generateEvent(`new base URI ${newBaseURI} well setted`);
+  } else {
+    generateEvent(`you are not the contract Owner`);
+  }
+  return '';
+}
+
+// ======================================================== //
+// ====                 TOKEN ATTIBUTES                ==== //
```
Thykof:
```suggestion
// ====                 TOKEN ATTRIBUTES                ==== //
```

assembly/NFT/__tests__/NFT.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    transfer(args);
+    const got = Storage.getOf(ofAddress, 'ownerOf_2');
+    const want = transferAddress.toByteString();
+    if (got != want) {
+      error(got.toString() + ', ' + want.toString() + ' was expected.');
+      return;
+    }
+  });
+  test('Change URI', () => {
+    setURI('my.massa/');
```
Thykof, line 103: put `my.massa/` in a constant variable

assembly/NFT/NFTWrapper.ts
```diff
@@ -0,0 +1,123 @@
...
+  tokenURI(tokenId: u64): string {
+    return this._baseURI + tokenId.toString();
+  }
+
+  /**
+   * Return the base URI (external link written in NFT where pictures or others a stored)
+   * @return {string}
+   */
+  baseURI(): string {
+    return this._baseURI;
```
Thykof, line 68: There might be an error here because if the function `setURI` is called (by the owner), then this function will return the previous value because the call to `baseURI` is only done in the constructor for the wrapper.

```suggestion
    this._baseURI = call(this._origin, 'baseURI', '', 0);
    return this._baseURI;
```

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+ *The NFT's main characteristics
+ */
+
+const _name: string = 'MASSA_NFT';
+const _symbol: string = 'NFT';
+const _maxSupply: string = '10000';
+const _baseURI: string = 'massa.net/nft/';
+
+/**
+ * Init the NFT with name, symbol, maxsupply and baseURI,
```
Thykof:
```suggestion
 * Init the NFT with name, symbol, maxSupply and baseURI,
```

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+const baseURIKey: string = 'baseURI';
+const initCounter: u64 = 0;
+
+/**
+ *The NFT's main characteristics
+ */
+
+const _name: string = 'MASSA_NFT';
+const _symbol: string = 'NFT';
+const _maxSupply: string = '10000';
```
Thykof, line 23: why do we hard code the max supply? Does it means that the user will not be able to deploy an NFT with a custom max supply?

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+const _baseURI: string = 'massa.net/nft/';
+
+/**
+ * Init the NFT with name, symbol, maxsupply and baseURI,
+ *
+ * init the counter to 0, the owner of the contract,
```
Thykof:
```suggestion
 * Init the counter to 0, the owner of the contract and the base URI.
```

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+  if (Storage.has(baseURIKey)) {
+    return Storage.get(baseURIKey);
+  } else {
+    return '';
+  }
+}
+
+/**
+ * Return the max supply possible
+ *@param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
```
Thykof:
```suggestion
 * @param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
```

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
...
+    const addr = Address.fromByteString(args);
+    _increment('');
+    const tokenID: string = currentSupply('');
+    const key = ownerTokenKey + tokenID;
+    Storage.set(key, addr.toByteString());
+    generateEvent(`tokenId ${tokenID} minted to ${addr.toByteString()} `);
+  } else {
+    generateEvent(`Max supply reached`);
+  }
+  return 'Minted';
```
Thykof, line 166: why does it returns something? and in any case?

Thykof (changes requested):

assembly/NFT/NFT_fortest.ts
```diff
@@ -0,0 +1,237 @@
...
+////////////////////////
+////////////////////////////
+///////////////////////////////
+
+// as astester.imports.js for the
+//  moment only contents setOf, getOf and generateEvent,
+// this NFT_fortest contract has been modificated in consequences
+// original contract kept commented to further implementation
+
+import {Address, Storage, Context, generateEvent} from '../std';
```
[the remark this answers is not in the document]
SebastienValla: kept to have it ready when uncommented code if mock Context function is implemented in astester.imports.ts,
If we remove this we should remove all commented code as well

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,232 @@
+///////////////////////////////
+////////////////////////////
+////////////////////////
+// WIP allowance, and transferFrom Missing
```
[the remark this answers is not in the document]
SebastienValla: no, there is not

assembly/NFT/NFT.ts
SebastienValla: "The operand of an increment or decrement operator must be a variable or a property access i implement" this

assembly/NFT/NFT.ts
SebastienValla: you just have to change it in the NFT contract

Trottoir (approved):

assembly/NFT/NFT.ts
Thykof: let's create an issue about this

assembly/NFT/NFTWrapper.ts
```diff
@@ -0,0 +1,121 @@
...
+  tokenURI(tokenId: u64): string {
+    return this._baseURI + tokenId.toString();
+  }
+
+  /**
+   * Return the base URI (external link written in NFT where pictures or others a stored)
+   * @return {string}
+   */
+  baseURI(): string {
+    return call(this._origin, 'baseURI', '', 0);
```
Thykof, line 68: As I have suggested here https://github.com/massalabs/massa-as-sdk/pull/28#discussion_r1016769927
tokenURI will return the old _baseURI if we do not update here.

assembly/NFT/__tests__/NFT.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      error(got.toString() + ', ' + want.toString() + ' was expected.');
+      return;
+    }
+  });
+  test('Change URI', () => {
+    const newURI = 'my.massa/';
+    setURI(newURI);
+    const got = baseURI('');
+    const want = 'my.massa/';
+    if (got != want) {
```
Thykof, line 110:
```suggestion
    const newURI = 'my.massa/';
    setURI(newURI);
    const got = baseURI('');
    if (got != newURI) {
```
My previous comment was not clear, Don't Repeat Yourself ;)

Or even:

```suggestion
    const newURI = 'my.massa/';
    setURI(newURI);
    if (baseURI('') != newURI) {
```

Thykof (changes requested):

Thykof (approved):

package.json
```diff
@@ -22,8 +22,8 @@
     "registry": "https://registry.npmjs.org/"
   },
   "scripts": {
-    "test": "asp --verbose",
-    "test:ci": "asp --summary",
+    "test:asp": "asp --verbose",
```
peterjah, line 25: what the point of adding ":asp" ?

assembly/NFT/NFT_fortest.ts
```diff
@@ -0,0 +1,237 @@
+///////////////////////////////
```
peterjah, line 1: Sorry but i think it is pointless to run tests on a modified copy of your code like this... If things are missing then we have to add it. But for me this workaround is not acceptable

peterjah (changes requested):

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * Init the NFT with name, symbol, maxSupply and baseURI,
+ * init the counter to 0, the owner of the contract.
+ * @param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
+ *
+ * @return {string}
+ */
+export function setNFT(_: string): string {
+  if (!Storage.has(counterKey)) {
+    Storage.set(baseURIKey, _baseURI);
```
peterjah, line 35: why you don't use setOf /getOf / hasOf ?

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+  return '';
+}
+
+/**
+ * Increment the NFT counter
+ * @param {string} _ - unused see https://github.com/massalabs/massa-sc-std/issues/18
+ * @return {string}
+ */
+function _increment(_: string): string {
```
peterjah, line 174: i read the issue here: https://github.com/massalabs/massa-sc-std/issues/18
But doest it really apply here ?? i mean its an internal function of the contract. it will not be called from the outside

package.json
Thykof: because in this repo, we have tests written for as-pect test framework and the new NFT contract test is written for envy test framework.

assembly/NFT/NFT.ts
Thykof: good catch, indeed this function is not exported

assembly/NFT/NFT.ts
Thykof: https://github.com/massalabs/massa-as-sdk/issues/32

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,229 @@
...
+const baseURIKey: string = 'baseURI';
+const initCounter: u64 = 0;
+
+/**
+ * The NFT's main characteristics
+ */
+
+const _name: string = 'MASSA_NFT';
+const _symbol: string = 'NFT';
+const _maxSupply: string = '10000';
```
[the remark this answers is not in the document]
SebastienValla: I made this choice to have uniformity in variables, if everything is a string you need to handle it on the same way.

assembly/NFT/NFT.ts
SebastienValla: why would you like store data in another address instead of the nft's ones ?

assembly/NFT/NFT.ts
```diff
@@ -0,0 +1,229 @@
...
+// ==================================================== //
+
+/**
+ * The to address becomes the owner of the next token (if current tokenID = 10, will mint 11 )
+ * Check if max supply is not reached
+ * @param {string} args - byte string containing an owner's account (Address).
+ * @return {string}
+ */
+export function mint(args: string): string {
+  if (u32(parseInt(limitSupply(''))) > u32(parseInt(currentSupply('')))) {
```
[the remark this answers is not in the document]
SebastienValla: see previous comment

package.json
SebastienValla: I will let you create a new issue for this

Thykof (approved):

peterjah (approved):

Thykof (approved):

PR #30, opened 2022-11-09
Replace gas price with fee

gterzian:
Part of https://github.com/massalabs/massa/pull/3173#discussion_r1014817564

### DIFF
--- a/assembly/std/index.ts
+++ b/assembly/std/index.ts
@@ -235,7 +235,7 @@ export function unsafeRandom(): i64 {
  * @param {u64} validityEndPeriod - Period of the validity end slot
  * @param {u8} validityEndThread - Thread of the validity end slot
  * @param {u64} maxGas - Maximum gas for the message execution
- * @param {u64} gasPrice - Price of one gas unit
+ * @param {u64} rawFee - Fee to be paid for message execution
  * @param {u64} coins - Coins of the sender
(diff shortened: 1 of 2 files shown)


assembly/std/index.ts
```diff
@@ -258,7 +258,7 @@ export function sendMessage(
     validityStartThread,
     validityEndPeriod,
     validityEndThread,
-    maxGas,
+ rawFees,
```
Thykof, line 261: You can use `npm run fmt` to format the code and see linting issues about jsdoc in order to make the CI green.

Thykof (changes requested):

Thykof (approved):

gterzian:
Should we merge this? What is the procedure for that? @Thykof

Thykof:
@gterzian I am asking @gregLibert for review

gregLibert (approved):

gregLibert:
thanks @gterzian !

qdrn:
This change will only be effective in the testnet 17. This makes the latest version of massa-as-sdk not compatible with the current testnet (used for hackathon). Is that a problem ? @gregLibert @Thykof

gregLibert:
Yes it is. We need to reverse...

gterzian:
Ok, restored the branch and will re-open later when we can merge...

PR #39, opened 2022-11-16
Change all the required types from str to byte array

Eitu33:
⚠️ DO NOT merge this until corresponding updates have been done on the other repositories


assembly/std/index.ts
```diff
@@ -32,8 +32,10 @@ export function call(
   functionName: string,
   args: Args,
   coins: u64,
-): string {
-  return env.call(at.toByteString(), functionName, args.serialize(), coins);
+): StaticArray<u8> {
+  // IMPORTANT TODO: change this after `Args` update
```
sydhds, line 36: Args feature is now ok no?

Eitu33: Yes but it does not serialize into the right data type, need to sync with the inno team on that to see if they want me to take care of it or if they'd rather do it themselves

AurelienFT: Done.

AurelienFT:
@ASAPSegfault and @peterjah I have put you as a review because you are aware of the situation but if you want feel free to switch it to someone else.

peterjah (approved):

PR #42, opened 2022-11-16
fixes toU32 length check

Thykof:


assembly/std/arguments.ts
```diff
@@ -198,7 +198,7 @@ export class Args {
    * @return {u32}
    */
   private toU32(byteArray: Uint8Array, offset: u8 = 0): u32 {
-    if (byteArray.length - offset < 8) {
+    if (byteArray.length - offset < 4) {
```
gregLibert, line 201: I guess that you should clarify the meaning of that value ;-) 
```suggestion
    if (byteArray.length - offset < sizeof<u32>()) {
```

gregLibert (changes requested):
🙅 we should remove the magic number that is at the origin of a bug and this correction :-).

We could also do that for u64 .

thomas-senechal (approved):

PR #43, opened 2022-11-16
Improve Args

### LINKED ISSUE #35: Improve Args class
Following #23 

High priority:
- [x] implement `add<f64>`, `add<f32>`, `nextF64` and `nextF32`

Lower priority:
- [x] transform into Uint8Array directly onto the constructor: make the private class variable serialized a type Uint8Array
- [x] add methods nextUint8Array and addUint8Array
- [x] in method AddAddress: transform the address into Uint8Array
- [x] in method nextAddress: create an Address from the Uint8Array
- [ ] optimize string serialization, see https://github.com/massalabs/massa-as-sdk/pull/23#discussion_r1015098969

Question: this class is also used to serialize data to put into storage's hash map, should we rename `Args` class into something like `Serializer`?

AurelienFT:
Fix #35


assembly/std/arguments.ts
```diff
@@ -127,28 +181,57 @@ export class Args {
    */
   add<T>(arg: T): Args {
     if (arg instanceof Address) {
-      this.serialized = this.serialized.concat(arg.toStringSegment());
+      let length = new Uint8Array(1);
+      length[0] = arg._value.length;
```
Thykof, line 185: you should use `arg.toStringSegment()` instead of arg._value.

I think here you need `this.add<u32>(arg.toStringSegment());`. unless the length is auto converted to uint8.

assembly/std/arguments.ts
```diff
@@ -170,6 +253,104 @@ export class Args {
...
+
+  /**
+   * Converts a f64 in a bytearray.
+   *
+   * @param {f64} number the number to convert
+   *
+   * @return {Uint8Array} the converted bytearray
+   */
+  private fromF64(number: f64): Uint8Array {
+    log(number);
```
Thykof, line 279: debug log here

assembly/std/arguments.ts
```diff
@@ -127,28 +181,57 @@ export class Args {
...
+      this.serialized = this.concatArrays(
+        this.serialized,
+        this.fromU32(changetype<u32>(arg)),
       );
     } else if (arg instanceof i64) {
-      this.serialized = this.serialized.concat(
-        ByteArray.fromI64(arg as i64).toByteString(),
+      this.serialized = this.concatArrays(
+        this.serialized,
+        this.fromU64(changetype<i64>(arg)),
```
Thykof, line 212: here you give i64 to function fromU64

Thykof (changes requested):

AurelienFT:
@Thykof I have resolved the conflicts.

AurelienFT:
Need https://github.com/massalabs/massa-as-sdk/pull/50 to be merged before.

Thykof (approved):

AurelienFT:
Merged in #52

Issue #29, opened 2022-11-08
Modify envy to be able tu use it from massa-as-sdk


SebastienValla:
In envy code there is if path=node_modules/@massalabs/massa-as-sdk/astester.imports.ts  do something else do other things,
so astester.imports.ts cannot be imported directly.

Thykof:
related to #25

gregLibert:
Discontinued initiative

PR #48, opened 2022-11-16
20 fungible token vote

0xMazout:
Adding : 

Cast Vote : To spend voting power on a proposal

CRUD Proposal : To manage proposal

Delegate : To Delegate our voting power

Storing datas : 
VotingData : Related to Voting Data proposal , 
Data : Related to Data proposal,
VotingData*ProposalID*AddrSpender* : Related to voting Data user on proposal
proposal : Related to all proposals created by the addr


0xMazout:
There is package.json push to correct the bug we find on asprettier windows stuff

0xMazout:
Currently creating some units tests but we can review the architecture  and the choice made

assembly/token/impl.ts
```diff
@@ -379,3 +381,603 @@ export function transferFrom(stringifyArgs: string): string {
...
+  const amount = args.nextU64();
+
+  if (!ownerAddress.isValid() || !recipientAddress.isValid() || isNaN(amount)) {
+    return '0';
+  }
+
+  const ownerBalance = _balance(ownerAddress);
+  const recipientBalance = _balance(recipientAddress);
+  
+  const delegatedPowerRecipient = new Args(Storage.getOf(recipientAddress, 'DelegatedPower')).nextU64();
```
SebastienValla, line 442: why use args in this case and not  u64(string) ?

assembly/token/impl.ts
```diff
@@ -379,3 +381,603 @@ export function transferFrom(stringifyArgs: string): string {
...
+    return 'Voting : Invalid reason';
+  }
+
+  // Check if the proposal State is Active
+  if (Storage.hasOf(proposalOwner, 'Data'.concat(proposalId))) {
+    let dataProposal = Storage.getOf(proposalOwner, 'Data'.concat(proposalId));
+    let argsproposal = new Args(dataProposal);
+    const owner = argsproposal.nextAddress();
+    const id = argsproposal.nextString();
+    const title = argsproposal.nextString();
```
SebastienValla, line 500: title is never used

SebastienValla (changes requested):

Thykof:
fix #29

Thykof:
this will be done in https://github.com/massalabs/massa-sc-contracts/pull/12

Issue #53, opened 2022-11-18
Use a library for base64 encode


Thykof:
We can use this library https://www.npmjs.com/package/as-base64 and remove base64.ts file from our code

gregLibert:
@Thykof It was a temporary work around. The TestNet 17 will remove the string in favor of a byte array.

PR #61, opened 2022-11-21
adds as-base64

Thykof:
Test fail with: 

```
ERROR TS6054: File '~lib/as-base64.ts' not found.
   :
 2 │ import {encode, decode} from 'as-base64';
   │                              ~~~~~~~~~~~
   └─ in assembly/std/arguments.ts(2,30)
```


Thykof:
close because it doesn't work and won't fix: https://github.com/massalabs/massa-as-sdk/issues/53#issuecomment-1322241489

Issue #44, opened 2022-11-16
Address struct not fully bytes encoded


Trottoir:
The `Address` struct : 
{
      _value : "string",
      _isValid : "bool"
}

All fields of the struct Address are not encoded through `toByteString`, only the _value field is encoded.

We need to also encode the validity of the address

gregLibert:
We don't. Validity should be checked at caller level.

PR #62, opened 2022-11-21
feat (args): serialize Address.isValid

Thykof:


assembly/std/address.ts
```diff
@@ -70,8 +70,8 @@ export class Address implements Valider {
    *
    * @return {Address}
    */
-  static fromByteString(bs: string): Address {
-    return new Address(bs);
+  static fromByteString(bs: string, isValid: bool = true): Address {
```
SebastienValla, line 73: why does it return always true ?

SebastienValla (changes requested):

assembly/std/address.ts
Thykof: not always true, the default value of the argument is true, is it was before but I need a way to change it to false during de-serialization

SebastienValla (approved):

assembly/std/arguments.ts
```diff
@@ -45,8 +45,10 @@ export class Args {
    */
   nextAddress(): Address {
     const length = this.nextU32();
+    const isValid = this.nextBool();
     let address = Address.fromByteArray(
       this.serialized.slice(this.offset as i32, (this.offset as i32) + length),
```
thomas-senechal, line 50: Since you changed offset to an `i32`, I think you can remove the `as i32`s from this line. And if there are other instances of `this.offset as i32`, I think the `as i32` can also be removed.

assembly/std/arguments.ts
```diff
@@ -16,7 +16,7 @@ import {encode, decode} from './base64';
  *
  */
 export class Args {
-  private offset: i64 = 0;
+  private offset: i32 = 0;
```
thomas-senechal, line 19: Is there a specific reason to change this variable from an `i64` to `i32` ?

thomas-senechal (changes requested):

assembly/std/arguments.ts
Thykof: because the length of a string is a i32: https://www.assemblyscript.org/stdlib/string.html#fields

assembly/std/arguments.ts
Thykof: good catch, thanks!

thomas-senechal (approved):

Ben-Rey:
What is the status of this PR ?

Thykof:
> What is the status of this PR ?

we can close the PR: https://github.com/massalabs/massa-as-sdk/issues/44#issuecomment-1333340448

PR #71, opened 2022-11-25
Aargs: serialize boolean

Thykof:


.editorconfig
```diff
@@ -5,7 +5,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
peterjah, line 8: 🎉

assembly/std/arguments.ts
```diff
@@ -46,7 +46,7 @@ export class Args {
   nextAddress(): Address {
     const length = this.nextU32();
     let address = Address.fromByteArray(
-      this.serialized.slice(this.offset as i32, (this.offset as i32) + length),
+      this.serialized.slice(this.offset, this.offset + length),
```
peterjah, line 49: 🎉

peterjah (approved):

assembly/std/__tests__/arguments.spec.ts
```diff
@@ -29,6 +29,16 @@ describe('Args tests', () => {
     expect(args2.nextU64()).toBe(97);
   });
```
SebastienValla, line 31: is it a good idea to stay on aspect ?

Thykof: not at all but we have an issue for migrating all aspect tests

SebastienValla (approved):

PR #78, opened 2022-11-28
Initial code for getKeys()

sydhds:


assembly/std/index.ts
```diff
@@ -139,6 +139,16 @@ export function getOpKeys(): Array<StaticArray<u8>> {
...
 }
 
+/**
+ * Get all keys from datastore
+ *
+ * @return {StaticArray<u8>} - a list of key (e.g. a list of bytearray)
+ */
+export function getKeys(address: string): Array<StaticArray<u8>> {
+  let keysSer = env.getKeys(address);
+  return derOpKeys(keysSer);
```
AurelienFT, line 149: Why to you use `derOpKeys` for this also? This function works in this case also I assume. Then I think you should rename it to show that it works on multiple usage.

sydhds: Renamed

AurelienFT (approved):

PR #80, opened 2022-11-28
79 change filetob64 function for new transformer

SebastienValla:


Thykof (approved):
LGTM

assembly/std/index.ts
```diff
@@ -266,21 +266,21 @@ export function sendMessage(
 }
 
 /**
- * Convert given file content to base64.
+ * Convert given file content to byteArray.
  *
  * Note: this function shall never be called but is dynamically
- * replace using base64 transformer.
+ * replace using byteArray transformer.
  * More info here:
```
peterjah, line 273: more info where ? :)

peterjah (approved):

assembly/std/index.ts
SebastienValla: do you want to put the transformer file folder here ? I don't think it make sense in case it changes

PR #82, opened 2022-11-28
add timestamp function

SebastienValla:


Thykof (approved):
LGTM

0xMazout (approved):
Tested & Approved 👍

astester.imports.js
```diff
@@ -172,6 +172,9 @@ export function local(memory) {
         ledger.set(newAddress._value, newAddressLedger);
         return newString(newAddress._value);
       },
+      assembly_script_get_time(){
+        return BigInt(Date.now());
```
gregLibert, line 176: You should force the conversion to an u64 value yourself.

```suggestion
        return BigInt.asUintN(64, Date.now());  
```

gregLibert (changes requested):
:mag: you should handle the proper length of your returned arguments type yourself.

gregLibert (approved):
:+1:

PR #87, opened 2022-11-30
Allow to use string & StaticArray<u8> for values and keys

Trottoir:


assembly/std/storage.ts
```diff
@@ -20,28 +33,43 @@ export function set(key: StaticArray<u8>, value: StaticArray<u8>): void {
  * TODO: explains security mecanisms
  *
  * @param {Address} address
- * @param {StaticArray<u8>} key
- * @param {StaticArray<u8>} value
+ * @param {T } key
```
Thykof, line 36:
```suggestion
 * @param {T} key
```

assembly/std/storage.ts
```diff
@@ -20,28 +33,43 @@ export function set(key: StaticArray<u8>, value: StaticArray<u8>): void {
  * TODO: explains security mecanisms
```
Thykof, line 33:
```suggestion
 * TODO: explains security mechanisms
```

Thykof (changes requested):
2 tiny suggestions.

Warning: those functions will now accept any types, so if you provide i32 for example, it will take `StaticArray<u8>(0)`. This change the behavior because before, passing i32 would not compile

assembly/std/storage.ts
```diff
@@ -6,11 +7,23 @@ import {Address} from './address';
...
  */
-export function set(key: StaticArray<u8>, value: StaticArray<u8>): void {
-  env.set(key, value);
+export function set<T>(key: T, value: T): void {
+  let keyByte =
+    key instanceof String
+      ? toBytes(key.toString())
+      : key instanceof StaticArray<u8>
+      ? (key as StaticArray<u8>)
+      : new StaticArray<u8>(0);
```
gregLibert, line 19: This code is too complex and dangerous: if you call set with something else than a string or a byte array, it will compile but failed at the execution because you will send an empty array for the key and the value.

The FFI concept is complex enough and I don't want to build magic stuff that doesn't work 100% of the time over it.

gregLibert (changes requested):
:recycle: this need should be covered in another way.

assembly/std/storage.ts
Trottoir: Yeah I agree with what you say, didn't notice that.

And what about if we abort when a key OR value is not a string or a byte array ? 

The other solution would be to build specific functions : 
- setOfString 
- getString 
....
But we'll multiply the amount differents functions exposed by the storage interface

gregLibert: On this specific subject, @Thykof had a very good answer here : https://discord.com/channels/721472913886281818/1047897635018780682

Trottoir:
Just applied the modifications from Discord, I'm just not sure about the "abort" function to throw the error.
Is that will revert fully the tx ?

gregLibert:
I have expanded this issue in order to manage both:
- strings,
- args, and
- staticArray<u8>

@Thykof can you have a look at it?

assembly/std/storage.ts
```diff
@@ -1,61 +1,131 @@
...
 /**
  * Returns (key, value) in the datastore of the given address.
  *
  * TODO: explains what happens on missing key.
  *
+ * @template {string|Args|StaticArray<u8>} T - string, Args or StaticArray<u8>
  * @param {Address} address
- * @param {StaticArray<u8>} key
+ * @param {T} key
  *
```
Thykof, line 119:
```suggestion
```

assembly/vm-mock/storage.ts
```diff
@@ -10,7 +12,7 @@
  */
 // @ts-ignore: decorator
 @external("massa-sc-std", "set")
```
Thykof, line 14: massa-sc-std is deprecated, what is the purpose of this annotation @external

Thykof (approved):

assembly/vm-mock/storage.ts
gregLibert: ok, this file can be deleted then

assembly/std/__tests__/storage.spec.ts
```diff
@@ -0,0 +1,42 @@
+import {Storage} from '..';
+import {Args} from '../arguments';
+
+describe('Storage tests', () => {
+  it('checks the set command', () => {
+    Storage.set(new StaticArray<u8>(0), new StaticArray<u8>(0));
+    Storage.set('1', '1');
+    Storage.set(new Args(), new Args());
+
+    // TODO: Define an implement how to test that.
```
Thykof, line 10:
```suggestion
    // TODO: Define and implement how to test that.
```

assembly/std/__tests__/storage.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    // Storage.set(new Uint8Array(0), new Uint8Array(0)); // doesn't compile with user error as expected.
+  });
+
+  it('checks the get command', () => {
+    Storage.get(new StaticArray<u8>(0));
+    Storage.get('1');
+    Storage.get(new Args());
+
+    // TODO: Define an implement how to test that.
```
Thykof, line 19:
```suggestion
    // TODO: Define and implement how to test that.
```

assembly/vm-mock/storage.ts
Thykof: Only setData seems to be not used, but others are used for example here: https://github.com/massalabs/massa-sc-contracts/blob/main/assembly/erc20/__test__/erc20-mint.spec.ts#L14

gregLibert (approved):
:tada: Well done!

assembly/std/storage.ts
```diff
@@ -1,61 +1,130 @@
...
+
+  // Not necessary, but when giving an unsupported type, avoid
+  // `ERROR TS2355: A function whose declared type is not 'void' must return a value.`
+  // which can be misleading when you try to figure out what caused the error.
+  return new StaticArray<u8>(0);
+}
+
+/**
+ * Converts given datastore retreived value to wanted format.
+ *
```
SebastienValla, line 37:
```suggestion
 * Converts given datastore retrieved value to wanted format.
 *
```

assembly/std/storage.ts
```diff
@@ -64,10 +133,11 @@ export function getOf(address: Address, key: StaticArray<u8>): StaticArray<u8> {
  * TODO: explains what happens on missing key.
  * TODO: explains security mecanisms
```
SebastienValla, line 134:
```suggestion
 * TODO: explains security mechanisms
```

assembly/std/storage.ts
```diff
@@ -76,11 +146,12 @@ export function del(key: StaticArray<u8>): void {
...
```
SebastienValla, line 147:
```suggestion
 * TODO: explains security mechanisms
```

assembly/std/storage.ts
```diff
@@ -103,40 +175,43 @@ export function append(key: StaticArray<u8>, value: StaticArray<u8>): void {
  * Note: do nothing if key is absent.
  * TODO: explains security mecanisms
```
SebastienValla, line 176:
```suggestion
 * TODO: explains security mechanisms
```

vm.js
```diff
@@ -0,0 +1,207 @@
...
+        const newAddress = {_value: generateDumbAddress(), _isValid: true};
+        const newAddressLedger = {
+          storage: new Map(),
+          contract: '',
+        };
+        ledger.set(newAddress._value, newAddressLedger);
+        return webModule.__newArrayBuffer(newAddress._value);
+      },
+      assembly_script_get_time(){
+        return BigInt.asUintN(64, Date.now());
```
SebastienValla, line 196:
```suggestion
        return BigInt(Date.now);  
```

vm.js
```diff
@@ -0,0 +1,207 @@
+/* eslint-disable require-jsdoc */
```
SebastienValla, line 1: Nice to did it for aspect !

SebastienValla (changes requested):

vm.js
Thykof: I will rebase on main

SebastienValla (approved):

PR #91, opened 2022-12-06
improve doc of stack addresses

Thykof:


damip:
@Thykof can you also create a documentation page with the whole explanation on how the stack works ?

SebastienValla (approved):

gregLibert:
> @Thykof can you also create a documentation page with the whole explanation on how the stack works ? (not just the documentation of those 3 functions)

@damip don't you think this documentation should be on Massa's read the doc ?

assembly/std/context.ts
```diff
@@ -37,7 +37,7 @@ export function ownedAddresses(): Array<Address> {
  * You get all previously called since the address of
  * the originator (transaction creator).
  *
- * @return {Array<Address>}
+ * @return {Array<Address>} returns the full call stack as a list, bottom to top
```
gregLibert, line 40: this bottom to top is unclear to me. What is top ? What is bottom ?

I propose:
```typescript
/*
 * Returns the stack of called addresses.
 * 
 * The first element of this list is the originator (creator of the transaction) and the last is the address of the current smart-contract.
 *
 * When executing the ExecuteSC function, the list is composed of only one entry: the initiator, the caller and the callee having with the same address.
 * 
 * @return {Array<Address>}
```

assembly/std/context.ts
```diff
@@ -49,7 +49,7 @@ export function addressStack(): Array<Address> {
  * Caller is the person or the smart contract that directly called
  * the pending function.
  *
- * @return {Address}
+ * @return {Address} returns the stack element just below the top
```
gregLibert, line 52: still not clear to me.

assembly/std/context.ts
```diff
@@ -61,7 +61,7 @@ export function caller(): Address {
  *
  * Callee is the smart contract of the pending function.
```
gregLibert, line 62:
```suggestion
 * Callee is the current smart-contract address.
```

Maybe we should create an alias named `currentAddress`

assembly/std/context.ts
```diff
@@ -71,7 +71,7 @@ export function callee(): Address {
 /**
  * Return the address of the initial transaction creator (originator).
  *
- * @return {Address}
+ * @return {Address} returns the stack element at the bottom of the stack
```
gregLibert, line 74: Same. Top/bottom...

gregLibert (changes requested):
This documentation is still unclear to me...

Thykof:
@gregLibert here is the doc https://docs.massa.net/en/latest/web3-dev/smart-contracts/address-stack.html

gregLibert (approved):
:clap: :camping:

0xMazout (commented):
Nice work

PR #132, opened 2023-01-05
add mock for sc call

peterjah:


assembly/vm-mock/env.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ *  Set mock for smartcontract call. When called multiple time, mocked values will be added in the mock stack.
+ *
+ * @param value - Value returned by the mocked function
+ */
```
gregLibert, line 5: Can you add some documentation explaining that this function expects an array of returned values?
Maybe an example?

It might also be interesting to explain that the values are returned sequentially.

vm-mock/vm.js
```diff
@@ -211,7 +213,21 @@ export default function createMockedABI(memory, createImports, instantiate, bina
...
+      assembly_script_mock_call(ptr) {
+        scCallMockStack.push(getArrayBuffer(ptr));
+      },
+
+      assembly_script_call(_address, method, _param, _coins) {
+
+        if(scCallMockStack.length) {
+          return newArrayBuffer(scCallMockStack.shift());
+        }
+        console.warn(`Warning: No mock defined for sc call to "${ptrToString(method)}". Returning empty data`);
```
gregLibert, line 227: Why not panicking ?

gregLibert (changes requested):

assembly/vm-mock/env.ts
peterjah: ok for the example.. but this does not expect an array. just a value to be used as return of the mock

vm-mock/vm.js
peterjah: its another valid strategy! and i have nothing against it. so lets go for it !

assembly/vm-mock/env.ts
peterjah: > just a value to be used as return of the mock

which is an array 😄

vm-mock/vm.js
```diff
@@ -211,7 +213,19 @@ export default function createMockedABI(memory, createImports, instantiate, bina
...
+
+      assembly_script_mock_call(ptr) {
+        scCallMockStack.push(getArrayBuffer(ptr));
+      },
+
+      assembly_script_call(_address, method, _param, _coins) {
+        if(scCallMockStack.length) {
+          return newArrayBuffer(scCallMockStack.shift());
+        }
+        throw new Error(`Warning: No mock defined for sc call on "${ptrToString(method)}".`);
```
Thykof, line 226:
```suggestion
        throw new Error(`No mock defined for sc call on "${ptrToString(method)}".`);
```

Thykof (approved):

SebastienValla (approved):
Nice job !

Thykof (approved):

gregLibert (approved):
:clap:

PR #140, opened 2023-01-12
improve doc and sendMessage

Thykof:


SebastienValla (approved):
LGTM !

assembly/std/index.ts
```diff
@@ -316,10 +316,30 @@ export function unsafeRandom(): i64 {
...
+ * "as soon as possible in this range" it becomes
+ * "as soon as possible after this field has been update in the state in this range".
+ *
+ * So you can pass as parameter a `filterAddress` and also an optional datastore `filterKey`.
+ *
+ * If you pass only an address then the message will be executed only after:
+ * "we are in the range and a change has happens during this range on the `filterAddress`" (possibly balance etc).
+ * If you provide a `filterKey` the condition of the execution of the message is:
+ * "we are in the range and a change has happens during this range on the `filterAddress` at that datastore `filterKey`"
+ *
```
longgamma-massa, line 338:
```suggestion
 * The goal of sendMessage functionality is to send a message in the future, that will be executed as soon as possible
 * after the start period but not after the end period. This message allows you to make executions in the future and they
 * are executed deterministically on all nodes. The execution is made "as soon as possible" because there is a priority
 * on messages and a limit of messages possibly executed on each slot. More precisely, if you send a low amount of `rawFee` then your
 * message could possibly not executed directly at the first slot of the slot period.
 * If all of the slots of the specified period have a large load of messages, with more fees than you specified, then it's highly likely that your
 * message will never be executed.
 *
 * Additionally, there is an optional filter on a message that adds a new condition on the trigger instead of:
 * "as soon as possible in this range", it becomes
 * "as soon as possible, after this field has been updated, in the state, in this range".
 *
 * As a parameter, you can pass the `filterAddress`, and also an optional datastore `filterKey`.
 *
 * If you pass only an address, then the message will be executed only after:
 * "we are in the range, and a change has happened during this range on the `filterAddress`" (possibly balance etc).
 * If you provide a `filterKey`, the condition of the execution of the message is:
 * "we are in the range, and a change has happened during this range on the `filterAddress` at that datastore `filterKey`"
 *
```

longgamma-massa: I have fixed the language but the whole paragraph needs more clarification. I suggest a small call so that we specify how to properly introduce a usecase or a few of them, before we start technical description of how it works

longgamma-massa (changes requested):

assembly/std/index.ts
Thykof: cc @AurelienFT

AurelienFT: Some usecases : 

- Having a smart contract called periodically without a centralized bot
- Having a smart contract that trigger on a value change of an other one (for example a price change)
- Having a object that evolve on the blockchain himself

longgamma-massa: Cool, thanks it will improve this section @AurelienFT .

assembly/std/index.ts
```diff
@@ -316,10 +316,31 @@ export function unsafeRandom(): i64 {
 /**
  * Sends an async message to a function at given address.
  *
+ * @remarks
+ * The goal of sendMessage functionality is to send a message in the future, that will be executed as soon as possible
+ * after the start period but not after the end period. This message allows you to make executions in the future and
```
longgamma-massa, line 321: After the first sentence: "The goal of sendMessage functionality is to send a message in the future, that will be executed as soon as possible after the start period but not after the end period." include:
 
You might want to use the sendMessage functionality:

- Having a smart contract called periodically, without a centralized bot;
- Having a smart contract that will trigger on the change of value (for example a change in price), of an other one;
- Having an object that evolves on the blockchain itself.

Continue with the rest of the content "This message allows you".....

longgamma-massa (changes requested):
after this change it's ready for the merge

longgamma-massa (approved):

PR #145, opened 2023-01-24
serialize Address

Thykof:


Thykof:
*marked as blocked by https://github.com/massalabs/as/issues/89*

assembly/std/address.ts
```diff
@@ -1,10 +1,10 @@
-import { Valider, ByteArray } from '@massalabs/as-types';
+import { Valider, Serializable, Args, Result } from '@massalabs/as-types';
 
 /**
  * A Massa's blockchain address.
  *
  */
-export class Address implements Valider {
+export class Address implements Valider, Serializable {
```
gregLibert, line 7: Oh shit Valider still exist ? Maybe we should mark it a deprecated.

assembly/std/address.ts
```diff
@@ -29,81 +29,55 @@ export class Address implements Valider {
...
+    }
+    this._value = result.unwrap();
+    return new Result(args.offset);
   }
 
   /**
    * Serialize to byte string.
    */
-  toByteString(): string {
+  toString(): string {
```
gregLibert, line 63: is it a big breaking change ?

gregLibert (approved):

assembly/std/address.ts
Thykof: no because only the name of the function changes, not the logic

assembly/std/address.ts
Thykof: okay, issue create here: https://github.com/massalabs/as/issues/92

peterjah (approved):

PR #163, opened 2023-03-07
add prefix param to getKeys and getKeysOf functions

Ben-Rey:


Thykof (changes requested):
It could be great to document prefix param in assembly/std/index.ts and in the same time document params of derKeys function

AurelienFT:
@litchipi The PR associated to your work on massa :)

assembly/std/index.ts
```diff
@@ -219,35 +219,43 @@ export function getOpKeys(): Array<StaticArray<u8>> {
  *
```
Thykof, line 219: prefix param description is missing, can you explain how work the prefix filter?

Thykof (changes requested):
it could be nice to have a better explanation of what derKeys is doing

assembly/env/env.ts
```diff
@@ -43,10 +43,13 @@ export namespace env {
   export declare function createSC(bytecode: StaticArray<u8>): string;
 
   @external("massa", "assembly_script_get_keys")
-  export declare function getKeys(): StaticArray<u8>;
+  export declare function getKeys(prefix?: string): StaticArray<u8>;
```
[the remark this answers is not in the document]
Ben-Rey: Thanks, I will add a default value. I choose string because address in assembly_script_get_keys_for is also a string and in my opinion it is weird on our side to have the address as a string and the prefix in StaticArray<u8>.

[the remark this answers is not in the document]
Ben-Rey: ok just saw sc-runtime. I'll put StaticArray

AurelienFT (approved):

assembly/std/index.ts
```diff
@@ -215,57 +215,86 @@ export function getOpKeys(): Array<StaticArray<u8>> {
...
+  let keysSer = env.getKeysOf(address, prefix);
   return derKeys(keysSer);
 }
 
 /**
- * Internal function - used by getOpKeys
+ * Read the number of keys from Uint8Array little endian format
+ *
+ * @param arr - Uint8Array
+ * @remarks - array length must be equal to 4
```
Thykof, line 251: Have you `npm run doc` to see the rendering of a remarks bellow a param?

"array length must be equal to 4" 
does it means:
- it is a fixed-length array
OR
- the length of the array is encoded with 4 bits?

Thykof (changes requested):

assembly/std/index.ts
Ben-Rey: Thanks for your comments, you are right, it was not clear. I changed the code and the comment a bit to make it more understandable. I am adding tests for get_keys_for and getKeys in vm.js and vm-mock.spec because they are missing.

Thykof (approved):

assembly/std/index.ts
Thykof: good job!

PR #164, opened 2023-03-08
implement validate address

Ben-Rey:


vm-mock/vm.js
```diff
@@ -5,8 +5,9 @@ const { createHash } = await import('node:crypto');
  */
 
 // Those both addresses have been randomly generated
-const callerAddress = 'A12UBnqTHDQALpocVBnkPNy7y5CndUJQTLutaVDDFgMJcq5kQiKq';
-const contractAddress = 'A12BqZEQ6sByhRLyEuf0YbQmcF2PsDdkNNG1akBJu9XcjZA1eT';
+const callerAddress = 'AU12UBnqTHDQALpocVBnkPNy7y5CndUJQTLutaVDDFgMJcq5kQiKq';
+const contractAddress = 'AS12BqZEQ6sByhRLyEuf0YbQmcF2PsDdkNNG1akBJu9XcjZA1eT';
+const addressBytesLength = 53;
```
Thykof, line 10: I though that addresses are not fix-sized. If it true and if here you intentionally simplified by fixing the length to 53, then it's fine, just add a comment to explain this.

Thykof (changes requested):

vm-mock/vm.js
Ben-Rey: In that case it was just for me to check something. @AurelienFT what would be a true test case, I did not understand how it works in the rust?

AurelienFT (approved):

vm-mock/vm.js
AurelienFT: The address is PREFIX+base58encode(varint version, bytes)

Thykof (approved):

vm-mock/vm.js
AurelienFT: It's not always 53

PR #189, opened 2023-03-30
update storage documentation

gregLibert:


assembly/std/storage.ts
```diff
@@ -188,49 +287,71 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
- * @privateRemarks
- * TODO: explains failure consequences.
+ * @param bytecode - The bytecode to be set. It should be a `StaticArray<u8>`.
  *
- * @param bytecode -
+ * @throws
+ * - a runtime error if the callee's address does not correspond to a smart contract in the ledger.
+ * - a runtime error if the caller lacks write permissions on the callee's address.
+ * - a runtime error if the callee's address is the same as the creator's address.
+ * - a runtime error if an error occurs while updating the bytecode in the speculative ledger.
```
gregLibert, line 335: @Ben-Rey I'm not sure about this part

assembly/std/storage.ts
```diff
@@ -188,49 +287,71 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
+ * @param address - The target address whose bytecode will be set. It should be of type `Address`.
+ * @param bytecode - The bytecode to be set for the `address`. It should be a `StaticArray<u8>`.
  *
- * @param address - target address -
- * @param bytecode -
+ * @throws
+ * - a runtime error if the `address` does not correspond to a smart contract in the ledger.
+ * - a runtime error if the caller lacks write permissions on the `address`.
+ * - a runtime error if the `address` is the same as the creator's address.
+ * - a runtime error if an error occurs while updating the bytecode in the speculative ledger.
```
gregLibert, line 354: @Ben-Rey I'm not sure about this part

assembly/std/storage.ts
```diff
@@ -188,49 +287,71 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
 }
 
 /**
- *  Sets the executable bytecode of the given address.
+ * Sets the executable bytecode of the given `address`.
  *
- * @privateRemarks
- * TODO: explains security mechanisms.
+ * @remarks
+ * TODO: (add issue) explains security mechanisms.
```
gregLibert, line 345: @Ben-Rey I guess it's the same thing than xxxOf functions no ?

gregLibert:
@Ben-Rey I extracted the storage part because the review is to big to do...

gregLibert:
@Ben-Rey the fix you propose here https://github.com/massalabs/massa-as-sdk/blob/3a63190dbd54964f8ac12aae76eab8c8ed7e63bd/assembly/tsdoc.json#L4-L7 block the code generation. When I apply it on my side, I don't have the storage namespace documnetation anymore...

Ben-Rey:
> @Ben-Rey the fix you propose here
> 
> https://github.com/massalabs/massa-as-sdk/blob/3a63190dbd54964f8ac12aae76eab8c8ed7e63bd/assembly/tsdoc.json#L4-L7
> 
> block the code generation. When I apply it on my side, I don't have the storage namespace documnetation anymore...

Found the problem

assembly/std/storage.ts
```diff
@@ -188,49 +287,71 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
+ * - if the caller is not authorized to modify the datastore.
+ * - [compilation only] if the `key` type are neither `string`, `StaticArray<u8>`, or `Args`.
  */
 export function hasOf<T>(address: Address, key: T): bool {
   return env.hasOf(address.toString(), toDatastoreFormat(key));
 }
 
 /**
- *  Sets the executable bytecode of the callee's address.
+ * Sets the executable bytecode of the current contract.
```
Ben-Rey, line 325: What is the bytecode and what would we need to set it ?

assembly/std/storage.ts
Eitu33: First 3 are right, last one is too but I'll add some specifications about the speculative ledger error. It can be one of:
1. address does not exist
2. bytecode size exceeds maximum allowed size
3. overflow on computing bytecode storage cost (should never happen because of 2.)

assembly/std/storage.ts
Eitu33: Same as https://github.com/massalabs/massa-as-sdk/pull/189#discussion_r1155646604

assembly/std/storage.ts
Eitu33: Bytecode is the raw representation of a WebAssembly file. To set it properly you need to have:
* a smart contract using the right AS std (which maps the massa node v.x.x execution VM ABIs state)
* compile this sc into a WASM file
* use the byte array representation of that file

Eitu33: I'm not sure that answers your question let me know if it's not the case

assembly/std/storage.ts
Ben-Rey: Perfect thanks

assembly/std/storage.ts
Ben-Rey: Thanks for the answer, do you have a use case where we would set the bytecode for the curent address ?

assembly/std/storage.ts
```diff
@@ -1,12 +1,64 @@
...
+ *
+ * The supported value types for the datastore include `string`, `Args` and `StaticArray<u8>`.
+ *
+ * This module also provides functions for setting the executable bytecode of a smart contract address.
+ *
+ * @remarks
+ * The datastore is used to store data that is expected to persist between contract executions, such as contract
+ * state or user information.
+ *
+ * The [set](../functions/Storage.set.html), [get](../functions/Storage.get.html),
```
Thykof, line 13: You can use @link 

like here https://github.com/massalabs/as/blob/main/packages/as-types/assembly/serialization/arrays.ts#L13

see doc here https://typedoc.org/tags/link/

try something like this:

```typescript
 * The {@link set}, {@link get},
```

assembly/std/storage.ts
Eitu33: From within an SC that would want to update itself for `setBytecode`. Or in a deployement SC to update the bytecode of an address that was created in the same scope (case where another address has ownership over it) for `setBytecodeOf`

assembly/std/storage.ts
```diff
@@ -188,49 +282,71 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
+ * See [related issue](https://github.com/massalabs/massa-as-sdk/issues/182)
+ *
+ * @param address - The target address whose bytecode will be set. It should be of type `Address`.
+ * @param bytecode - The bytecode to be set for the `address`. It should be a `StaticArray<u8>`.
  *
- * @param address - target address -
- * @param bytecode -
+ * @throws
+ * - if the `address` does not correspond to a smart contract in the ledger.
+ * - if the `address` is the same as the creator's address.
```
Eitu33, line 346: This line is wrong, this is not a check we perform

assembly/std/storage.ts
```diff
@@ -188,49 +282,71 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
+ * Sets the executable bytecode of the current address.
  *
- * @privateRemarks
- * TODO: explains failure consequences.
+ * @param bytecode - The bytecode to be set. It should be a `StaticArray<u8>`.
  *
- * @param bytecode -
+ * @throws
+ * - if the caller address does not correspond to a smart contract in the ledger.
+ * - if the bytecode size exceeds maximum allowed size.
```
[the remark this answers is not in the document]
Ben-Rey: I removed these check because I thought:
- If a smart contract update itself it should have an address. 
- Because it update itself, how could it lack permissions.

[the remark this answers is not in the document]
Ben-Rey: Am I missing something ?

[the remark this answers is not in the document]
Ben-Rey: I will add them to reflect the core.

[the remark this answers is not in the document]
Eitu33: You can still call `set_bytecode` from an SC that is not stored on the ledger, in that case the caller address will not correspond to any address in the ledger. And yes you are right about the permissions, if it does not fail for preemptive reasons this error should not happen when its called from a stored SC.

assembly/std/storage.ts
Ben-Rey: A lot cleaner thanks

assembly/std/storage.ts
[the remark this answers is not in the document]
Ben-Rey: 🙏

assembly/std/storage.ts
```diff
@@ -188,49 +280,72 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
+ * @typeParam T - Can be either `string`, `Args`, or `StaticArray<u8>`.
+ *
+ * @param address - The child smart contract `address`.
+ * @param key - The key to check for existence in the datastore.
+ *
+ * @returns A boolean value indicating whether the key-value pair exists in the datastore or not.
+ *
+ * @throws
+ * - if the contract at the given `address` does not exist
+ * - if the caller is not authorized to modify the datastore.
```
Thykof, line 312:
```suggestion
```

assembly/std/storage.ts
```diff
@@ -188,49 +280,72 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
 
 /**
- * Checks if the (key, value) exists in the datastore
- * of the given address.
+ * Checks if the key-value pair exists in the datastore of the target address.
  *
- * @typeParam T - `string`, `Args` or `StaticArray<u8>`
- * @param address -
- * @param key -
+ * @remarks Can only be called at smart contract generation time by the parent smart contract to the child's address.
```
Thykof, line 301: I think this function reads only

```suggestion
```

assembly/std/storage.ts
```diff
@@ -188,49 +280,72 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
+ * Sets the executable bytecode of the current address.
  *
- * @privateRemarks
- * TODO: explains failure consequences.
+ * @param bytecode - The bytecode to be set. It should be a `StaticArray<u8>`.
  *
- * @param bytecode -
+ * @throws
+ * - if the caller address does not correspond to a smart contract in the ledger.
+ * - if the bytecode size exceeds maximum allowed size.
```
Thykof, line 326: As a developer, where can I found this maximum bytecode size?

assembly/std/storage.ts
```diff
@@ -188,49 +280,72 @@ export function appendOf<T>(address: Address, key: T, value: T): void {
...
+ * - if the address does not exist
+ * - if the caller lacks write permissions on the address.
```
Thykof, line 328: `Sets the executable bytecode of the current address`

As I understand, there is no address parameter involved, it will take Context.caller()

```suggestion
```

Thykof (changes requested):

assembly/std/storage.ts
Ben-Rey: we had a discussion about this here: https://github.com/massalabs/massa-as-sdk/pull/189#discussion_r1157043161

assembly/std/storage.ts
Ben-Rey: @Eitu33 can you help us on that ?

assembly/std/storage.ts
Ben-Rey: You're right thanks

assembly/std/storage.ts
Eitu33: `massa/massa-models/src/config/constants.rs` => `MAX_BYTECODE_LENGTH`

Ben-Rey: I added it in the tsDoc

PR #208, opened 2023-04-07
Documentation: Updated index module documentation.

0xSwapFeeder:


assembly/std/index.ts
```diff
@@ -305,21 +432,25 @@ export function derKeys(keysSer: StaticArray<u8>): Array<StaticArray<u8>> {
...
+ * This function converts the given string data into a base58 formatted string.
+ *
+ * @param data - the string data to convert to base58.
  *
- * @param data -
+ * @returns the converted data as a string.
  *
  */
 export function toBase58(data: string): string {
   return env.toBase58(data);
```
0xSwapFeeder, line 443: @BatiGencho This function seems to not exist in the env module is this normal ? Or should we remove this function ?

assembly/std/index.ts
```diff
@@ -75,63 +140,94 @@ export function localExecution(
...
  */
-export function createSC(bytecode: StaticArray<u8>): Address {
-  return new Address(env.createSC(bytecode));
+export function functionExists(address: Address, func: string): bool {
+  return env.functionExists(address.toString(), func);
 }
 
 /**
- * Generates an event
+ * Generates a string event that is then emitted by the blockchain and can be listened off-chain.
```
Ben-Rey, line 224: I thought about something simpler
```suggestion
 * Generate a log in the Massa blockchain
```

assembly/std/index.ts
```diff
@@ -75,63 +140,94 @@ export function localExecution(
...
+ *
+ * @see [massa smart-contracts examples](https://github.com/massalabs/massa-sc-examples) to see how to listen
```
Ben-Rey, line 226: The link is a bit confusing as it point to the massa-as-example repo. I would remove it

assembly/std/index.ts
```diff
@@ -140,28 +236,42 @@ export function generateEvent(event: string): void {
 /**
  * Transfers SCE coins from the current address to given address.
```
Ben-Rey, line 237: I would not use abreviation in function description. it can make harder for developers to understand the purpose of the function.

assembly/std/index.ts
```diff
@@ -140,28 +236,42 @@ export function generateEvent(event: string): void {
...
  *
- * @param to -
+ * @param to - the address to send coins to.
  * @param amount - value in the smallest unit.
```
Ben-Rey, line 240: We can explain what is the smallest unit in a remark.

see: [unit standard](https://github.com/massalabs/massa-standards/blob/main/units.md)
Or just a link. What do you think @gregLibert ?

assembly/std/index.ts
```diff
@@ -140,28 +236,42 @@ export function generateEvent(event: string): void {
...
+ * - if the balance of the current address is insufficient to make the transaction.
+ *
  */
 export function transferCoins(to: Address, amount: u64): void {
   env.transferCoins(to.toString(), amount);
 }
 
 /**
- * Transfers SCE coins of the `from` address to the `to` address.
+ * Transfers SCE coins 'from' a given address 'to' another given address.
```
Ben-Rey, line 252: Same here, we can remove SCE

assembly/std/index.ts
```diff
@@ -140,28 +236,42 @@ export function generateEvent(event: string): void {
...
+ *
+ * @remarks
+ * The transfer is done only after approval.
```
Ben-Rey, line 255: What do you mean by that ?

assembly/std/index.ts
Ben-Rey:
```suggestion
 * Transfer an amount of coins `from` the specified address `to` a target address.
```

assembly/std/index.ts
```diff
@@ -170,20 +280,25 @@ export function balance(): u64 {
 /**
  * Gets the balance of the specified address.
```
Ben-Rey, line 281:
```suggestion
 * Returns the balance of the specified address
```

assembly/std/index.ts
```diff
@@ -170,20 +280,25 @@ export function balance(): u64 {
...
  *
- * @param address -
+ * @param address - the address on which the balance is checked.
```
Ben-Rey, line 283:
```suggestion
 * @param address - The address for which the balance is retrieved.
```

assembly/std/index.ts
```diff
@@ -331,18 +461,28 @@ export function isSignatureValid(
...
  *
- * @param pubKey - Base58check encoded
  */
 export function publicKeyToAddress(pubKey: string): Address {
   return new Address(env.publicKeyToAddress(pubKey));
 }
 
 /**
- * Returns an unsafe random.
+ * Generates a pseudo-random integer.
```
Ben-Rey, line 479: I would keep unsafe
```suggestion
 * Generates an unsafe random integer.
```

assembly/std/index.ts
```diff
@@ -443,46 +585,61 @@ export function fileToByteArray(
...
+ * @returns the current period as u64.
+ *
  */
 export function currentPeriod(): u64 {
   return env.currentPeriod();
 }
 
 /**
- * Returns the current thread
+ * Retrieves the current thread of the network.
```
Ben-Rey, line 598: Returns the current thread where the contract is being executed

assembly/std/index.ts
```diff
@@ -443,46 +585,61 @@ export function fileToByteArray(
...
   return env.currentThread();
 }
 
 /**
- * Constructs an event given a key and arguments
+ * Constructs a pretty formatted event with given key and arguments.
+ * It is useful to generate events with array formatted data.
+ *
+ * @remarks
+ * The generated event is meant to be used with the {@link generateEvent} function.
```
Ben-Rey, line 612:
```suggestion
 * The createEvent is meant to be used with the {@link generateEvent} function.
```

assembly/std/index.ts
```diff
@@ -443,46 +585,61 @@ export function fileToByteArray(
...
+ * @param bytecode - the bytecode for which to compute the checksum.
+ *
+ * @returns the serialized computed checksum.
  *
- * @param bytecode - StaticArray<u8>
- * @returns - Computed Sha256 in StaticArray<u8>
  */
 export function sha256(bytecode: StaticArray<u8>): StaticArray<u8> {
   return env.sha256(bytecode);
 }
```
Ben-Rey, line 635:
```suggestion
/**
 * Computes the SHA256 hash of the given `data`.
 *
 * @remarks
 * The SHA256 hash algorithm produces a 32-byte hash, which is returned as a `StaticArray<u8>`.
 *
 * @param data - The data to hash.
 *
 * @returns The SHA256 hash of the `data`, serialized as a `StaticArray<u8>`.
 *
 */
export function sha256(data: StaticArray<u8>): StaticArray<u8> {
  return env.sha256(data);
}
```

assembly/std/index.ts
0xSwapFeeder: Transfering coins from an other address than the current one might be protected ?
Didn't found any infos in massa-sc-runtime.

assembly/std/index.ts
```diff
@@ -194,32 +303,38 @@ export function hasOpKey(key: StaticArray<u8>): bool {
...
 
 /**
- * Get all keys from operation datastore
+ * Retrieves all the keys from the operation datastore.
+ *
+ * @returns - a list of keys (e.g. a list of byte array)
  *
- * @returns - a list of key (e.g. a list of byte array)
  */
 export function getOpKeys(): Array<StaticArray<u8>> {
```
0xSwapFeeder, line 326: Since the function getKeys and getKeysOf are doing the exact same has this function but with optional parameter in addition:  is this function to be marked deprecated ?

assembly/std/index.ts
```diff
@@ -331,18 +458,28 @@ export function isSignatureValid(
...
+ * @param pubKey - Base58check encoded public key of the address
+ *
+ * @returns the fetched address as an 'Address' object.
+ *
+ * @throws
+ * - if the public key is invalid
  *
- * @param pubKey - Base58check encoded
  */
 export function publicKeyToAddress(pubKey: string): Address {
```
0xSwapFeeder, line 471: Can this function go in the Address class as a static function ?

assembly/std/index.ts
```diff
@@ -443,46 +582,64 @@ export function fileToByteArray(
...
+ * @param data - The data to hash.
+ *
+ * @returns The SHA256 hash of the `data`, serialized as a `StaticArray<u8>`.
  *
- * @param bytecode - StaticArray<u8>
- * @returns - Computed Sha256 in StaticArray<u8>
  */
-export function sha256(bytecode: StaticArray<u8>): StaticArray<u8> {
-  return env.sha256(bytecode);
+export function sha256(data: StaticArray<u8>): StaticArray<u8> {
```
0xSwapFeeder, line 633: Is this function fully generic ?
Even if so, was it built to answer a specific use case to describe to users ?
Such as computing the hash of a contract ?

assembly/std/index.ts
```diff
@@ -443,46 +582,64 @@ export function fileToByteArray(
...
- * @param address - Address to check
- * @returns boolean - true if the address is valid, false otherwise
+ * Checks if the given address is valid.
+ *
+ * @param address - the string address to validate.
+ *
+ * @returns 'true' if the address is valid, 'false' otherwise.
  *
  */
 export function validateAddress(address: string): bool {
```
0xSwapFeeder, line 645: Can this function migrate to the Address class as a static function ?

assembly/std/index.ts
Ben-Rey: @AurelienFT How do we know we have the right to use assembly_script_transfer_coins_for. For contract that we own ?

assembly/std/index.ts
Ben-Rey: For now we can leave it like that.

assembly/std/index.ts
Ben-Rey: For now we can leave it like that. There is an issue to implement validate_address in Address object

assembly/std/index.ts
Ben-Rey: We can live with this for now as it could be a clue when reading the smart contract code that we want the actual keys.

assembly/std/index.ts
```diff
@@ -7,45 +50,60 @@ import { Args } from '@massalabs/as-types';
...
+ * - if the function doesn't exist in the contract to call.
  *
- * @returns function returned value
  */
 export function call(
   at: Address,
   functionName: string,
   args: Args,
-  coins: u64,
+  coins: u64 = 0,
```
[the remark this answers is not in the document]
Ben-Rey: Since this is a critical value, I think it's best not to set a default value and leave it to the developer of the smart contract.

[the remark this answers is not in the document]
0xSwapFeeder: Should I rephrase the description to make sure the user understands the argument is not optional but it can take 0 as value ?

assembly/std/index.ts
AurelienFT: No, I think this could be done directly on AS side. We don't need massa info for this and so passing to massa-sc-runtime will cost gas for nothing

assembly/std/index.ts
AurelienFT: You can only use this endpoint when you have the write access on the address you are passing as parameter. This is possible at SC address contract creation for example

Side note: SCE balance do no exist anymore they have been merged together since few months there is only one balance now.

assembly/std/index.ts
gregLibert: Yes, you can reference the standard in a remark!

assembly/std/index.ts
0xSwapFeeder: Since this issue is meant for documentation, should we open another one to implement this then ?

Ben-Rey: Yes good idea.

assembly/std/index.ts
```diff
@@ -37,15 +87,21 @@ export function call(
 }
 
 /**
- * Calls a remote function located at given address within the current context.
+ * Calls a function from a remote contract in the current context.
```
gregLibert, line 90: Does it mean that the remote function is executed in the local context  ?

gregLibert (approved):
One simple question. Nothing blocking the merge.

Good job boys!

assembly/std/index.ts
0xSwapFeeder: Yes, you want me to add 'current local context' instead ?

PR #209, opened 2023-04-08
Update powered-by.yml

qdrn:
I've added the corresponding secret and updated the branch protection rule to allow bot to bypass it. It should fix the CI.


qdrn:
Ping me if it doesn't work/you need further help on things that you cannot do due to restricted rights.

gregLibert (approved):

Issue #190, opened 2023-03-31
 organize and explain organization


gregLibert:
This package is a mess, we should:

in a not destructive way (duplicate and use deprecated tag)

 ## Organize the functions ...
... in consistent modules

- [x] #253
- [x] #242
- [x] #222 - extract execution functions into dedicated modules 

## Rename functions ...
... based on there usage, not necessarly how it works under the hood on the rust side and with simpler terms.

- [ ] #252

## Document all that

- [x] #257
- [x] for each module
- [x] update the tsdoc dedicated readme file

0xSwapFeeder:
#222 Aim's to tackle the module organization part.

gregLibert:
> #222 Aim's to tackle the module organization part.

Can you update the initial issue instead ? Check the boxes that are covered and for each line check, add a trailing `- #<issue number>` to let the reader know wher we are on this subject

gregLibert:
@Ben-Rey can we close this ?

PR #212, opened 2023-04-12
211 Add VM external Storage related functions

Ben-Rey:


assembly/__tests__/env-storage.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  test('Testing del (assembly_script_delete_data)', () => {
+    const key: string = 'test';
+    const value: string = 'value';
+    // given
+    Storage.set(key, value);
+    expect(Storage.get(key)).not.toBeNull();
+    // when
+    env.del(stringToBytes(key));
+    // then
+    expect(Storage.get(key)).toBeFalsy();
```
gregLibert, line 26: This will panic if executed in a smart contract. You should :
- panic if the get key doesn't exist
- use has in you code instead to match your test

assembly/__tests__/env-storage.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+const testAddress = new Address(
+  'AU12E6N5BFAdC2wyiBV6VJjqkWhpz1kLVp2XpbRdSnL1mKjCWT6oR',
+);
+
+beforeEach(() => {
+  resetStorage();
+});
+
+describe('Testing env storage functions', () => {
+  test('Testing del (assembly_script_delete_data)', () => {
```
gregLibert, line 17: Feels weird when reading this:
- test + Testing
- nobody care/need of assembly_script_delete_data because it's an internal function

I have two suggestions:
```suggestion
  test('delete of a key', () => {
```
or
```suggestion
  it('deletes a key', () => {
```

assembly/__tests__/env-storage.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    env.del(stringToBytes(key));
+    // then
+    expect(Storage.get(key)).toBeFalsy();
+  });
+
+  test('Testing deleteOf (assembly_script_delete_data_for)', () => {
+    const key: string = 'test';
+    const value: string = 'value';
+    // given
+    Storage.setOf(testAddress, key, value);
```
gregLibert, line 33: What is testAddress ? Current address ? In that case you should call it like that because if not, we shouldn't be able to set the address of another address except if we are deploying it. but thiss scenario would be to much for unit test I guess.

assembly/__tests__/env-storage.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(Storage.getOf(testAddress, key)).not.toBeNull();
+    // when
+    env.deleteOf(testAddress.toString(), stringToBytes(key));
+    // then
+    expect(Storage.getOf(testAddress, key)).toBeFalsy();
```
gregLibert, line 38: Same as above, this will b^panic on a live node

assembly/env/env.ts
```diff
@@ -74,7 +74,7 @@ export namespace env {
   ): StaticArray<u8>;
 
   @external("massa", "assembly_script_delete_data")
-  export declare function del(key: StaticArray<u8>): void;
+  export declare function del(key: StaticArray<u8>): StaticArray<u8>;
```
gregLibert, line 77: Are you sure ?

vm-mock/vm.js
```diff
@@ -279,10 +279,66 @@ export default function createMockedABI(
           const addressStorage = ledger.get(contractAddress).storage;
           if (addressStorage.has(k)) {
             addressStorage.delete(k);
+          } else {
+            console.log('key not found');
```
gregLibert, line 283: Maybe it will panic in that scenaior. The [documentation](https://as-sdk.docs.massa.net/functions/Storage.del.html) is not clear on that.

vm-mock/vm.js
```diff
@@ -279,10 +279,66 @@ export default function createMockedABI(
...
+      assembly_script_delete_data_for(addressPtr, keyPtr) {
+        const address = ptrToString(addressPtr);
+        const key = ptrToUint8ArrayString(keyPtr);
+
+        if (ledger.has(address)) {
+          const addressStorage = ledger.get(address).storage;
+          if (addressStorage.has(key)) {
+            addressStorage.delete(key);
+          } else {
+            console.log('key not found');
```
gregLibert, line 297: same (panic ?)

vm-mock/vm.js
```diff
@@ -279,10 +279,66 @@ export default function createMockedABI(
...
+          }
+        } else {
+          console.log('address not found');
```
gregLibert, line 300: same (panic ?)

vm-mock/vm.js
```diff
@@ -279,10 +279,66 @@ export default function createMockedABI(
...
+        }
+      },
+
+      assembly_script_append_data(keyPtr, valuePtr) {
+        const address = contractAddress;
+        const key = ptrToUint8ArrayString(keyPtr);
+        const newValue = byteArrToUTF8String(getArrayBuffer(valuePtr));
+
+        if (!ledger.has(address)) {
+          console.log('address not found');
```
gregLibert, line 310:
```suggestion
          console.log('address not found');
          return;
```

vm-mock/vm.js
```diff
@@ -279,10 +279,66 @@ export default function createMockedABI(
...
+        }
+      },
+
+      assembly_script_append_data_for(addressPtr, keyPtr, valuePtr) {
+        const address = ptrToString(addressPtr);
+        const key = ptrToUint8ArrayString(keyPtr);
+        const newValue = byteArrToUTF8String(getArrayBuffer(valuePtr));
+
+        if (!ledger.has(address)) {
+          console.log('address not found');
```
gregLibert, line 329:
```suggestion
          console.log('address not found');
          return;
```

gregLibert (changes requested):

assembly/__tests__/env-storage.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  test('Testing append (assembly_script_append_data)', () => {
+    const key: string = 'test';
+    const value: string = 'hello';
+    const appendValue: string = 'world';
+    // given
+    Storage.set(key, value);
+    expect(Storage.get(key)).toBe(value);
+    // when
+    env.append(stringToBytes(key), stringToBytes(appendValue));
```
Thykof, line 49: if env.append modifies the storage, why it is not in Storage namespace?

in addition, it could handle string type for its arguments (like Storage.set)

```suggestion
    Storage.append(key, appendValue);
```

same for appendOf

(maybe it's not in the scope of this PR)

Thykof (changes requested):

assembly/__tests__/env-storage.spec.ts
gregLibert: > if env.append modifies the storage, why it is not in Storage namespace?

reorganisation will be done in another task #190

Thykof: cool!

assembly/env/env.ts
Ben-Rey: don't know what happened here, thanks

assembly/__tests__/env-storage.spec.ts
Ben-Rey: I replaced by has(key) 
And throw an error in vm side

assembly/__tests__/env-storage.spec.ts
Ben-Rey: I corrected it as above

vm-mock/vm.js
Ben-Rey: I corrected it as above

vm-mock/vm.js
Ben-Rey: I throw an error instead to match runtime

vm-mock/vm.js
Ben-Rey: I corrected it as above

gregLibert (approved):

PR #214, opened 2023-04-12
213 mock coins related functions

Ben-Rey:


vm-mock/vm.js
```diff
@@ -376,6 +379,53 @@ export default function createMockedABI(
...
+      assembly_script_transfer_coins(_addressPtr, _coinsAmount) {
+        const address = ptrToString(_addressPtr);
+        if (!ledger.has(address)) {
+          ledger.set(address, {
+            storage: new Map(),
+            contract: '',
+            balance: 0,
+          });
+        }
+        ledger.get(address).balance += Number(_coinsAmount);
```
peterjah, line 392: i think balances should be in BigInt, isn't it ?

Ben-Rey: you're right! Done

gregLibert (approved):

PR #227, opened 2023-04-14
222 sdk new architecture

0xSwapFeeder:
Here is first iteration of the new sdk's architecture, feel free to suggest any changes that could be made.
(To see a better overview of the result of these changes, run:
```
npm run doc
```
And open the **doc/html/index.html** file in your web browser.


Ben-Rey:
@0xSwapFeeder You can remove the doc from index.ts  that has been moved to other files

assembly/std/coins/index.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * This namespace is used to interact with wallet's **balance** and **transfers**.
+ * It is located at **std/coins**.
```
gregLibert, line 3: I think we should mathc what the dev when loading the library is seeing instead of the location of the file.

hence I propose:
```suggestion
 * It is located at **coins**.
```

assembly/std/coins/transfer.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param to - the address to send coins to.
+ * @param amount - value in the smallest unit.
+ * @see [massa units standard](https://github.com/massalabs/massa-standards/blob/main/units.md)
+ *
+ * @throws
+ * - if the given address is not a valid address.
+ * - if the balance of the current address is insufficient to make the transaction.
+ *
+ */
+export function transferCoins(to: Address, amount: u64): void {
```
gregLibert, line 16: I propose to name it transfer as we are in the coins package:
```suggestion
export function transfer(to: Address, amount: u64): void {
```

assembly/std/coins/transfer.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param to - the address to send coins to.
+ * @param amount - value in the smallest unit.
+ * @see [massa units standard](https://github.com/massalabs/massa-standards/blob/main/units.md)
+ *
+ * @throws
+ * - if the sender's or the receiver address is not a valid address.
+ * - if the balance of the sender's address is insufficient to make the transaction.
+ *
+ */
+export function transferCoinsOf(from: Address, to: Address, amount: u64): void {
```
gregLibert, line 37: same:
```suggestion
export function transferOf(from: Address, to: Address, amount: u64): void {
```

assembly/std/contract/calls.ts
```diff
@@ -0,0 +1,192 @@
...
+  at: Address,
+  functionName: string,
+  args: Args,
+  coins: u64,
+): StaticArray<u8> {
+  return env.call(at.toString(), functionName, args.serialize(), coins);
+}
+
+/**
+ * Calls a function from a remote contract in the current context.
```
gregLibert, line 33:
```suggestion
 * Calls a function from a remote contract and execute it in the current context.
```

assembly/std/contract/index.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * This namespace is used to interact with other **smart contract functions** and manipulate their **bytecode**.
+ * It is located at **std/contract**.
```
gregLibert, line 3: same:
```suggestion
 * It is located at **contract**.
```

assembly/std/op-datastore/index.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * This namespace is used to interact with the **operation datastore**.
+ * It is located at **std/op-datastore**.
```
gregLibert, line 3:
```suggestion
 * It is located at **op-datastore**.
```

assembly/std/op-datastore/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @remarks
```
gregLibert, line 5: I guess it's a private remarks as it doesn't change the way the dev interact with theis library

assembly/std/utils/index.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * This namespace is a compilation of utilities functions relative to **smart contract development**.
+ * It is located at **std/utils**.
```
gregLibert, line 3: same:
```suggestion
 * It is located at **utils**.
```

assembly/std/utils/index.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @remarks
```
gregLibert, line 5: Private?

assembly/std/contract/index.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @remarks
```
gregLibert, line 5: private?

assembly/std/coins/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @remarks
```
gregLibert, line 5: private?

gregLibert (changes requested):
Well done!

Do you know if this will brake compatibility ? If yes, we need to publish a new version first and then update major digit.

Ben-Rey:
> Well done!
> 
> Do you know if this will brake compatibility ? If yes, we need to publish a new version first and then update major digit.

Normally, there is no major change. We export the functions from the index file, which gives the same result as before.

assembly/std/op-datastore/index.ts
0xSwapFeeder: The reason why I did this listing is for the exemple of the namespace 'Utilities'  which has many different sub-folders to categorize functions, so the purpose was to explain to the user theses implicit sub-namespaces, but maybe we should do theses explicitly to avoid such redundant function listing and have proper descriptions ?

assembly/std/coins/transfer.ts
0xSwapFeeder: To be done in another issue.

assembly/std/coins/transfer.ts
0xSwapFeeder: To be done in another issue.

assembly/std/op-datastore/index.ts
gregLibert: you're right.

PR #228, opened 2023-04-14
[fix] assembly_script_caller_has_write_access

Ben-Rey:
This is an error from conflict resolution

### DIFF
--- a/vm-mock/vm.js
+++ b/vm-mock/vm.js
@@ -398,7 +398,7 @@ export default function createMockedABI(
       },
 
       assembly_script_caller_has_write_access() {
-        return false;
+        return true;
       },
 
       isDeployingContract() {
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Massa-as-sdk
-![check-code-coverage](https://img.shields.io/badge/coverage-78%25-orange)
+![check-code-coverage](https://img.shields.io/badge/coverage-79%25-orange)
+
 
 
 


vm-mock/vm.js
```diff
@@ -398,7 +398,7 @@ export default function createMockedABI(
       },
 
       assembly_script_caller_has_write_access() {
-        return false;
+        return true;
```
gregLibert, line 401: Ar eyou sure ? I think that by default the caller should not have write access...

gregLibert (changes requested):

vm-mock/vm.js
Ben-Rey: it can be blocking for someone how want to test a constructor:

```
export function constructor(_: StaticArray<u8>): void {
  // This line is important. It ensures that this function can't be called in the future.
  // If you remove this check, someone could call your constructor function and reset your smart contract.
  if (!callerHasWriteAccess()) {
    return;
  }
  ```
  
Until we have a better way to test it, maybe it's best to return true.

wdyt?

gregLibert: I hnik that constructor should use isDeployingSC() instead and that be default, the unit test shouldn't be considering that function the smart contract is being deployed.

If I were you I would add:
- a function to set the value of isDeployingSC to let people test constructor
- leave that functin to false.

Ben-Rey:
will be fixed here: https://github.com/massalabs/massa-as-sdk/pull/243

Issue #229, opened 2023-04-14
Validate address


Ben-Rey:
**Describe the bug**
Currently validate address in vm-mock does not match the way the core validate an address.

Ben-Rey:
@sydhds @Leo-Besancon Do you have an idea on how I could validate a address in js and be as close as possible to the way it is done in rust ?

Leo-Besancon:
Hi @Ben-Rey! So I assume because it's regarding a vm-mock you can't just call an ABI :p. In that case I think you'll have to handle the first characters manually (checking AU / AS prefix), and then use a library to decode the base58check?
For example, the [bitcoinjs/bs58check](https://github.com/bitcoinjs/bs58check/blob/master/base.js) repo has a decode function that checks that the checksum of the data you want to decode is valid.

Once it's done, you still may need to verify the version bytes from decoded data.

Ben-Rey:
> Hi @Ben-Rey! So I assume because it's regarding a vm-mock you can't just call an ABI :p. In that case I think you'll have to handle the first characters manually (checking AU / AS prefix), and then use a library to decode the base58check? For example, the [bitcoinjs/bs58check](https://github.com/bitcoinjs/bs58check/blob/master/base.js) repo has a decode function that checks that the checksum of the data you want to decode is valid.
> 
> Once it's done, you still may need to verify the version bytes from decoded data.

Thanks !

PR #230, opened 2023-04-14
[fix] add range to address verification to match variable length

Ben-Rey:
This is a temporary fix. The reason is address can have different length so it's better for now to have a range. It will be completely fixed with : https://github.com/massalabs/massa-as-sdk/issues/229#issuecomment-1510793590

### DIFF
--- a/vm-mock/vm.js
+++ b/vm-mock/vm.js
@@ -414,11 +414,20 @@ export default function createMockedABI(
 
       assembly_script_validate_address(addressPtr) {
         const address = ptrToString(addressPtr);
+        const prefix = 'A';
+        const userOrSC = ['U', 'S'];
+        const minLength = 40;
+        const maxLength = 60;
 
-        return (
-          stringToByteArray(address).length === addressBytesLength &&
-          (address.startsWith('AU') || address.startsWith('AS'))
-        );
+        if (address.length < minLength && address.length > maxLength) {
+          return false;
+        }
+
+        if (address.startsWith(prefix) && userOrSC.includes(address[1])) {
+          return true;
+        }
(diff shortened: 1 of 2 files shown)


gregLibert (approved):

vm-mock/vm.js
```diff
@@ -414,11 +414,20 @@ export default function createMockedABI(
...
-        );
+        if (address.length < minLength && address.length > maxLength) {
+          return false;
+        }
+
+        if (address.startsWith(prefix) && userOrSC.includes(address[1])) {
+          return true;
+        }
+
+        return false;
```
Thykof, line 430: You can simplify the code like this

```suggestion
        return address.startsWith(prefix) && userOrSC.includes(address[1])
```

In addition and no-blocking, I like when the condition is self-explaining, so that:
- it improves readeability
- you can unit test each condition

```ts
  return isRightLength(address) && isRightPrefix(address)
```

```ts
function isRigthLength(address) {
  return address.length >= minLength || address.length <= maxLength
}

function isRightPrefix(address) {
  return address.startsWith(prefix) && userOrSC.includes(address[1]))
}
```

then in the following PR, you will just change isRigthLength implementation


```ts
function isRigthLength(address) {
  return base58 decode is valid...
}
```

Thykof (changes requested):

vm-mock/vm.js
Ben-Rey: great thanks! I'll change it

Ben-Rey: Done

Thykof (approved):

PR #243, opened 2023-04-18
Feature: Added mock context helpers functions

### LINKED ISSUE #241: Vm-mock context helpers for sc developpers
## **Context**
For now, sc developpers are not able to test their protected functions such as their **constructor** 
or even other function they want to protect from public usage.

## **User flow**
- The users should be able to mock an 'administrator like context' where they have write access on the contract they are testing.
- The users should be able to mock the 'deploy context' where they can interact with their contracts as they would be able to if they just deployed it with another contract.

## **How to**
I propose to implement some vm-mock helper functions:
- [x] 'mockAdminContext()' to give write access
- [x] 'mockDeployContext()' to give write access and emulate deployement
- [x] 'resetContext()' to reset both context

**Technical details**
To do this, we should expose some AS helper functions, implement their corresponding logic/states in the vm.js file
as well as update 'assembly_script_caller_has_write_access' and 'isDeployingContract' in vm.js to make a good use of the newly created states.

**QA testing**
To validate these functionnalities we should simply add test cases for both 'assembly_script_caller_has_write_access' and 'isDeployingContract' functions where we should be able to emulate their expected behavior in different context states set by theses new helpers.

0xSwapFeeder:
Aims to resolve #241 

Since isDeployContext is not an env function it was removed from vm.js mocker.

Added functions instead to modify the call stack so that the user can emulate both local and deploy context.


vm-mock/vm.js
```diff
@@ -427,12 +427,74 @@ export default function createMockedABI(
...
+      /**
+       * Sets the current context to a local context by setting the caller address to the contract address.
+       * Gives write access on any sc function tested.
+       * If the address is not set, uses the current contract address as caller address.
+       * 
+       * @param {*} addrPtr - the new optional address of the call stack
+       */
+      assembly_script_set_local_context(addrPtr) {
+        adminContext = true;
+        if (addrPtr == undefined || addrPtr == 0 || addrPtr == null) {
```
Ben-Rey, line 479:
```suggestion
        if (!addrPtr) {
```

assembly/vm-mock/env.ts
```diff
@@ -30,3 +30,92 @@ export declare function mockScCall(value: StaticArray<u8>): void;
...
+export declare function mockAdminContext(): void;
+
+/**
+ * Reset the context to the default one by removing all the write access
+ * and the deployment emulation.
+ *
+ * @example
+ * ```typescript
+ * test('test function protection', () => {
+ *   resetContext();
```
Ben-Rey, line 62: As I understand it should be mockNonAdminContext or maybe the function should be called resetContext()

vm-mock/vm.js
```diff
@@ -427,12 +427,62 @@ export default function createMockedABI(
...
+      assembly_script_mock_not_admin_context() {
+        adminContext = false;
+      },
+
+      assembly_script_set_deploy_context(addrPtr) {
+        adminContext = true;
+        // Ensure the caller address is different from the contract address
+        console.log("\n\ncallStack Before: " + callStack);
+        console.log("callerAddress Before: " + callerAddress);
+        console.log("contractAddress Before: " + contractAddress);
```
Ben-Rey, line 443: should we remove log ?

Ben-Rey (changes requested):

vm-mock/vm.js
0xSwapFeeder: absolutely !

gregLibert (approved):

PR #251, opened 2023-04-20
[CI] use only one version of node

Ben-Rey:


Ben-Rey:
will be done here https://github.com/massalabs/massa-as-sdk/pull/256/files

PR #261, opened 2023-04-20
fix unit-test CI

gregLibert:


gregLibert:
i have fixed the CI. Can you fix [the test](https://github.com/massalabs/massa-as-sdk/blob/main/assembly/std/__tests__/context.spec.ts#L17) ?

From my understanding, this is not working because [here](https://github.com/massalabs/massa-as-sdk/blob/main/assembly/__tests__/vm-mock.spec.ts#L196) and [here](https://github.com/massalabs/massa-as-sdk/blob/main/assembly/__tests__/vm-mock.spec.ts#L231) we are breaking the callstack. Maybe we should add an helper to restore the callstack as we are restoring the storage and to call it beforeAll test or after all test when we break it.

0xSwapFeeder:
> i have fixed the CI. Can you fix [the test](https://github.com/massalabs/massa-as-sdk/blob/main/assembly/std/__tests__/context.spec.ts#L17) ?
> 
> From my understanding, this is not working because [here](https://github.com/massalabs/massa-as-sdk/blob/main/assembly/__tests__/vm-mock.spec.ts#L196) and [here](https://github.com/massalabs/massa-as-sdk/blob/main/assembly/__tests__/vm-mock.spec.ts#L231) we are breaking the callstack. Maybe we should add an helper to restore the callstack as we are restoring the storage and to call it beforeAll test or after all test when we break it.

That's odd, it is already what does [breaking callstack tests](https://github.com/massalabs/massa-as-sdk/blob/main/assembly/__tests__/vm-mock.spec.ts#L196).

However it is not updating the global addresses, just the call stack, that might be it i'll check.

0xSwapFeeder (commented):
There is still an issue with the CI, when trying to update README.md, it doesn't have the permission @gregLibert

PR #267, opened 2023-04-26
[doc] improve localCall doc

Ben-Rey:


bilboquet (approved):
Thanks :)

assembly/std/contract/calls.ts
```diff
@@ -35,6 +35,12 @@ export function call(
  * @remarks
  * Arguments serialization is to be handled by the caller and the callee.
  *
+ * @privateRemarks
+ * When using localCall, the contract calls a function defined in another smart contract
+ * and executes it within its own context. This means that if a datastore key is modified
+ * in the called function, the modification will occur in the current smart contract's datastore.
+ * This behavior is attributed to the execution taking place "in the current context" of the caller.
```
gregLibert, line 42: this should be a public remarks, it will guide the dev using the library on what this function is doing.
```suggestion
 * When using localCall, the contract calls a function defined in another smart contract
 * and executes it within its own context. This means that if a datastore key is modified
 * in the called function, the modification will occur in the current smart contract's datastore.
 * This behavior is attributed to the execution taking place "in the current context" of the caller.
 *
 * Arguments serialization is to be handled by the caller and the callee.
```

gregLibert (approved):

PR #289, opened 2023-08-10
Add AbiEncode & friends + Bytes32 + Bytes4 classes

sydhds:


assembly/std/solidity_compat/bytes.ts
```diff
@@ -0,0 +1,103 @@
...
+      //        fromArray
+      memory.copy(
+        changetype<usize>(this.serialized) + this.offset_ser,
+        changetype<usize>(arg),
+        arg.length,
+      );
+      this.offset_ser += arg.length;
+    } else if (arg instanceof string) {
+      let _arg = stringToBytes(arg);
+      assert(this.offset_ser + arg.length <= this.MAX_LEN);
```
Ben-Rey, line 26:
```suggestion
      assert(this.offset_ser + _arg.length <= this.MAX_LEN);
```

assembly/std/solidity_compat/bytes.ts
```diff
@@ -0,0 +1,103 @@
...
+    return this;
+  }
+
+  public serialize(): StaticArray<u8> {
+    return this.serialized;
+  }
+
+  public deserialize(data: StaticArray<u8>, offset: i32 = 0): Result<i32> {
+    const args = new Args(data, offset);
+    this.serialized = args.getNextData(this.max_len());
```
Ben-Rey, line 89: getNextData is private

assembly/std/solidity_compat/bytes.ts
```diff
@@ -0,0 +1,103 @@
+import { Serializable, stringToBytes } from '@massalabs/as-types';
```
Ben-Rey, line 1:
```suggestion
import { Args, Result, Serializable, stringToBytes } from '@massalabs/as-types';
```

assembly/std/solidity_compat/bytes.ts
```diff
@@ -0,0 +1,103 @@
...
+export class Bytes32 {
+  private MAX_LEN: i32 = 32;
+  // private _offset: i32 = 0;
+  private serialized: StaticArray<u8> = new StaticArray<u8>(32);
+  private offset_ser: i32 = 0;
+
+  add<T>(arg: T): Bytes32 {
+    // let len = this.serialized.length;
+
+    if (arg instanceof StaticArray<u8>) {
```
Ben-Rey, line 12: I don't think you can use instanceof in assemblyscript

assembly/std/solidity_compat/bytes.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  abstract max_len(): i32;
+
+  constructor() {
+    this.serialized = new StaticArray<u8>(this.max_len());
+  }
+
+  _add<T>(arg: T): this {
+    const MAX_LEN = this.max_len();
+    if (arg instanceof StaticArray<u8>) {
```
Ben-Rey, line 56: same here.

assembly/std/solidity_compat/bytes.ts
Ben-Rey: My mistake

assembly/std/solidity_compat/bytes.ts
sydhds: Fixed

assembly/std/solidity_compat/bytes.ts
sydhds: Removed section so no need to pull in Args & Result

PR #290, opened 2023-08-16
PersistentMap: add support for StaticArray<u8> + Serilizable object

sydhds:


Ben-Rey (approved):

assembly/collections/__tests__/persistentMap.spec.ts
```diff
@@ -87,4 +88,61 @@ describe('Persistent Map tests', () => {
...
+      valueB,
+    );
+
+    // expect<Address>(map.getSome(keyC, new Address())).toStrictEqual(valueB);
+    // expect<Address | null>(map.get(keyC, new Address())).toStrictEqual(new Address());
+    expect<Address | null>(map.get(keyC, new Address())).toStrictEqual(
+      new Address(),
+    );
+    let res = map.getSome(keyC, new Address());
+    assert(res.isErr(), 'getSome for keyC should return an Result in error');
```
Ben-Rey, line 146: I don't understand  => 'getSome for keyC should return an Result in error'

sydhds: On line 145, there is a getSome called and it is expected to return a Result which should be an error variant Result

Ben-Rey, line 146:
```suggestion
    assert(res.isErr(), 'getSome for keyC should return a Result in error');
```

Ben-Rey (approved):

PR #311, opened 2023-11-23
Fix audit

Ben-Rey:


assembly/std/context.ts
```diff
@@ -62,8 +62,8 @@ export function isDeployingContract(): bool {
  *
  * @returns An array of `Address` objects, one for each address in the input string.
  */
-function json2Address(str: string): Array<Address> {
-  str = str.substr(1, str.length - 2);
+export function json2Address(str: string): Array<Address> {
```
peterjah, line 65: i really don't understand why this is in massa as std lib :/

peterjah (approved):

assembly/std/context.ts
Ben-Rey: You're right I could move it to an utils folder what do you think ?

PR #313, opened 2023-12-12
Add chainId external function

sydhds:


peterjah (approved):

Ben-Rey (approved):

Ben-Rey (approved):

vm-mock/vm.js
```diff
@@ -87,6 +87,7 @@ function resetLedger() {
 let webModule;
 
 const scCallMockStack = [];
+let chainIdMock = BigInt(77658366); // Default value, chain id for chain id
```
Thykof, line 90: maybe better like this?
`Default value, chain id for buildnet`

vm-mock/vm.js
```diff
@@ -774,6 +775,12 @@ export default function createMockedABI(
         const hash = sha3.keccak256.arrayBuffer(data);
         return newArrayBuffer(hash);
       },
+      assembly_script_set_chain_id(value) {
+        chainIdMock = value;
+      },
+      assembly_script_chain_id() {
+        return chainIdMock;
+      },
```
Thykof, line 783: I don't see unit test for this, are we sure it works? most of functions here handle params and return values as pointers

PR #315, opened 2023-12-18
Add chainId external function

Ben-Rey:
Reverts massalabs/massa-as-sdk#314

### DIFF
--- a/assembly/vm-mock/env.ts
+++ b/assembly/vm-mock/env.ts
@@ -19,6 +19,9 @@
 @external("massa", "assembly_script_mock_call")
 export declare function mockScCall(value: StaticArray<u8>): void;
 
+@external("massa", "assembly_script_set_chain_id")
+export declare function mockSetChainId(value: number): void;
+
 /**
  * Add a new smart contract address to the ledger
  *
--- a/vm-mock/vm.js
+++ b/vm-mock/vm.js
@@ -87,6 +87,7 @@ function resetLedger() {
 let webModule;
 
 const scCallMockStack = [];
+let chainIdMock = BigInt(77658366); // Default value, chain id for chain id
 
 /**
  * Create a mock vm to simulate calls and responses of Massa WebAssembly sdk.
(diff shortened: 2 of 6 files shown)


vm-mock/vm.js
```diff
@@ -87,6 +87,7 @@ function resetLedger() {
 let webModule;
 
 const scCallMockStack = [];
+let chainIdMock = BigInt(77658366); // Default value, chain id for chain id
```
Thykof, line 90:
```suggestion
let chainIdMock = BigInt(77658366); // Default value, chain id for BuildNet
```

Thykof:
see my previous comment here https://github.com/massalabs/massa-as-sdk/pull/313#discussion_r1430192373

Ben-Rey:
> see my previous comment here [#313 (comment)](https://github.com/massalabs/massa-as-sdk/pull/313#discussion_r1430192373)

@sydhds Could you provide a simple unit test to validate that it works?

vm-mock/vm.js
```diff
@@ -774,6 +775,12 @@ export default function createMockedABI(
         const hash = sha3.keccak256.arrayBuffer(data);
         return newArrayBuffer(hash);
       },
+      assembly_script_set_chain_id(value) {
+        chainIdMock = value;
+      },
+      assembly_script_chain_id() {
+        return chainIdMock;
```
peterjah, line 782: maybe you can return a default value, in case the mock as not been initialized?

assembly/vm-mock/env.ts
```diff
@@ -19,6 +19,9 @@
 @external("massa", "assembly_script_mock_call")
 export declare function mockScCall(value: StaticArray<u8>): void;
 
+@external("massa", "assembly_script_set_chain_id")
+export declare function mockSetChainId(value: number): void;
+
```
peterjah, line 24: i dont see real usecase where this would be needed.. 
i also don't see any usecase where chainId would be needed in a smart contract. It would mean you are developing a SC compatible with multiple chains ? and do some stuff differently on purpose?
sounds weird to me. and just not optimized in term of gas use

vm-mock/vm.js
sydhds: The default value is set to the chain id of the Buildnet

vm-mock/vm.js
sydhds: Fixed

assembly/vm-mock/env.ts
sydhds: Used by Umbrella SC for instance, closing this...

Ben-Rey:
Done here https://github.com/massalabs/massa-as-sdk/pull/316

PR #333, opened 2024-03-18
Add assertIsSmartContract function and tests

Ben-Rey:


vm-mock/vm.js
```diff
@@ -700,6 +705,15 @@ export default function createMockedABI(
...
         }
 
+        // We should not have this case, but someone could add a ledger.set without a balance
+        // Typescript could be used to avoid this kind of error
+        if (ledger.get(toAddress).balance === undefined) {
+          ledger.get(toAddress).balance = BigInt(0);
+        }
+        if (ledger.get(fromAddress).balance === undefined) {
+          ledger.get(fromAddress).balance = BigInt(0);
+        }
```
peterjah, line 715: the `ledger.get(fromAddress).balance` is already used previously in the function. Moreover if you set fromBalance to 0, the result will be a negative balance. so imho you should throw an error (the one wich is already throwed at line 705

assembly/__tests__/env-coins.spec.ts
```diff
@@ -52,6 +60,18 @@ describe('Testing env coins related functions', () => {
     expect(env.balanceOf(testAddress2.toString())).toBe(amount);
   });
 
+  it('transferCoins to mock address created with addAddressToLedger', () => {
```
peterjah, line 63: i find your test description quite hard to understand :/
It seems your test is just doing a simple transfercoins ? isn't it?

assembly/__tests__/env-coins.spec.ts
```diff
@@ -52,6 +60,18 @@ describe('Testing env coins related functions', () => {
...
+    addAddressToLedger(testAddress3.toString());
+    const amount: u64 = 100;
+    const receiverCurrentBalance = env.balanceOf(testAddress3.toString());
+    // given
+    expect(receiverCurrentBalance).toBe(0);
+    // when
+    env.transferCoins(testAddress3.toString(), amount);
+    // then
+    expect(env.balanceOf(testAddress3.toString())).toBe(amount);
```
peterjah, line 72: you should also test that sender balance decreased

assembly/__tests__/env-coins.spec.ts
Ben-Rey: Yes I'm not sure it's usefull now because we should not be able to create a new address with undifined balance

peterjah (approved):

PR #340, opened 2024-05-23
339 constant key manager

gregLibert:


Ben-Rey (approved):

leoloco (approved):
Shouldn't we implement some sort of domain separator for multiple files smart contracts ?
To prevent reentrancy/duplicate keys issues.
Or maybe it should be done at sc level

PR #341, opened 2024-05-23
Add is contract address

Ben-Rey:


assembly/std/address.ts
```diff
@@ -114,4 +115,33 @@ export class Address implements Serializable {
...
+
+  /**
+   * Checks if the address is a smart contract.
+   *
+   * @param address - The address to check.
+   *
+   * @returns `true` if the address is a smart contract, `false` otherwise.
+   *
+   * @throws
+   * If no bytecode is found at the address.
```
gregLibert, line 131: Not anymore.

assembly/std/address.ts
```diff
@@ -114,4 +115,33 @@ export class Address implements Serializable {
   notEqual(address: Address): boolean {
     return !(this == address);
   }
+
+  isValid(): bool {
```
gregLibert, line 119: You should add a description to be consistent with the rest

gregLibert (changes requested):

gregLibert (approved):

PR #343, opened 2024-05-24
Fix ConstantManager.set to accept array of serializable

gregLibert:


ASAPSegfault (approved):

ASAPSegfault:
LGTM But you need to check why unit tests are failing

gregLibert:
> LGTM But you need to check why unit tests are failing

They are failing because the fix on Args (as) is needed.

PR #352, opened 2024-06-06
add generate raw event function

Ben-Rey:


assembly/std/utils/arrays.ts
```diff
@@ -0,0 +1,11 @@
+export function staticArrayToUint8Array(
+  staticArray: StaticArray<u8>,
+): Uint8Array {
+  let uint8Array = new Uint8Array(staticArray.length);
+
+  for (let i = 0; i < staticArray.length; i++) {
+    uint8Array[i] = staticArray[i];
+  }
+
+  return uint8Array;
```
gregLibert, line 10: What about using something like:
```typescript
const arr = new Uint8Array(staticArray.length);
memory.copy(changetype<usize>(arr), changetype<usize>(staticArray), arr.length);
return arr;
```

gregLibert (changes requested):

assembly/std/utils/arrays.ts
Ben-Rey: I've moved the function in the project because it could be useful elsewhere.

gregLibert (approved):

PR #365, opened 2025-01-30
Features/solidity compat doc 1

sydhds:


assembly/std/solidity_compat/misc.ts
```diff
@@ -10,3 +15,25 @@ import { wrapStaticArray } from '@massalabs/as-types';
...
+export function selfDestruct(transferToAddr: Address): void {
+  // 1- empty the SC
+  let emptySc = new StaticArray<u8>(0);
+  env.setBytecode(emptySc);
+
+  // 2- delete everything in Storage
+  let keys = env.getKeys(new StaticArray<u8>(0));
+  for (let i = 0; i < keys.length; i++) {
+    Storage.del(keys[i]);
+  }
```
Ben-Rey, line 30: If to many Keys I think this function will fail. Could we add a system to batch key removal.

sydhds: I've added a small note for developers about that :)

Ben-Rey (approved):

assembly/std/solidity_compat/misc.ts
```diff
@@ -10,3 +15,27 @@ import { wrapStaticArray } from '@massalabs/as-types';
...
+// https://www.infuy.com/blog/using-self-destruct-in-solidity-contracts/
+// Note: this function will try to delete all Storage keys. Devs might want to clean things up manually
+//       before using this function.
+export function selfDestruct(transferToAddr: Address): void {
+  // 1- empty the SC
+  let emptySc = new StaticArray<u8>(0);
+  env.setBytecode(emptySc);
+
+  // 2- delete everything in Storage
+  let keys = env.getKeys(new StaticArray<u8>(0));
```
peterjah, line 29: I think you want to use `getKeys()` in storage.ts, not `env.getKeys`
`env.getKeys`return one single buffer altough `getKeys()` while "deserialize" the buffer into an array of keys

